PR #5, opened 2020-07-08
fix: refactor kustomization to fix releases

andrewrynhard:


config/kustomization.yaml
```diff
@@ -1,9 +1,12 @@
...
   - namespace.yaml
 
 bases:
   - ../internal/app/cluster-api-provider/config
   - ../internal/app/metal-controller-manager/config
   - ../internal/app/metal-metadata-server/config
+
+commonLabels:
+  app: sidero
+  cluster.x-k8s.io/provider: 'sidero'
```
andrewrynhard, line 11: This should be on every resource: https://github.com/kubernetes-sigs/cluster-api/issues/1989

PR #12, opened 2020-07-14
fix: update kustomize configs

rsmitty:
This PR updates the kustomize for the various sidero components. First,
it ensures that the namespace is set to sidero-system for each app.
Additionally, there are some RBAC changes that allow for reading of
secrets by the metadata server so it can lookup bootstrap data.


smira:
looks like DCO is missing

andrewrynhard:
@smira good catch. I will add conform to this.

andrewrynhard:
Added the webhook.

andrewrynhard:
We can force merge this, and I will follow up with the other conform fixes.

PR #19, opened 2020-08-07
docs: address markdown lint errors

andrewrynhard:
This fixes some lint errors, and allows long lines and inline HTML.


rsmitty:
Ah thank you. I knew there was a way to disable those and couldn't find it right away.

Issue #43, opened 2020-08-26
Create your first cluster guide


rsmitty:
What to do after bootstrapping and pivoting. Should be pretty basic.

PR #50, opened 2020-09-03
docs: add a post-pivot/create first workload cluster guide

### LINKED ISSUE #42: Guide for post-pivot
There are some things that need to be done after pivoting. First that comes to mind is editing the default environment to update the $PUBLIC_IP. Should take note of these and add them to the "creating your first cluster" guide.

rsmitty:
This PR adds a guide for what comes after the initial bootsrap and
pivot. I think I've got some other ideas on how this can get a bit
easier over time. I'll create a new PR with that later.

Will close #42 #43

### DIFF
--- a/docs/website/content/docs/bootstrapping.md
+++ b/docs/website/content/docs/bootstrapping.md
@@ -2,7 +2,7 @@
 description: "A guide for bootstrapping Sidero management plane"
 sidebar: "docs"
 prev: "/docs/installation/"
-next: "/docs/patching/"
+next: "/docs/first-cluster/"
 ---
 
 # Bootstrapping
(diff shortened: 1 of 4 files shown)


docs/website/content/docs/bootstrapping.md
```diff
@@ -254,7 +252,7 @@ Now that we have the manifest, we can simply apply it:
 kubectl apply -f management-plane.yaml
 ```
 
-> NOTE: The templated manifest above is meant to act as a starting point. If customizations are needed to ensure proper setup of your Talos cluster, they should be added before applying.
```
andrewrynhard, line 257: nit: The way we have written notes in our other sites is with the blockquote (`>`). We can merge your change, but one thing I have wanted to do for a long time now is create components that render notes in a "standard" way.

We should look at https://github.com/gridsome/gridsome/issues/677, and https://www.npmjs.com/package/@gridsome/vue-remark. Looks like we should be able to make components similar to: https://matcornic.github.io/hugo-learn-doc/cont/shortcodes/. 

e.g.

```markdown
<note>Some note.</note>
```

andrewrynhard (approved):

Issue #53, opened 2020-09-09
Nodes start booting image when environment is created


smira:
As soon as default environment is created, nodes start booting the kernel/initramfs from the environment, which is not expected to happen.

With Talos, luckily nodes can't fetch config from metadata server, but it might lead to other consequences for different server types.

PR #58, opened 2020-09-10
feat: add decommission state

andrewrynhard:
This adds a feature that automatically decommissions a server after being
deleted from a cluster.


app/metal-controller-manager/internal/server/server.go
```diff
@@ -84,7 +84,61 @@ func (s *server) CreateServer(ctx context.Context, in *api.CreateServerRequest)
...
+	// })
+	// if err != nil {
+	// 	if !apierrors.IsAlreadyExists(err) {
+	// 		return nil, err
+	// 	}
+
+	// 	log.Printf("%s already exists", in.GetSystemInformation().GetUuid())
+	// } else {
+	// 	log.Printf("Added %s", in.GetSystemInformation().GetUuid())
+	// }
```
andrewrynhard, line 137: WIP

app/metal-controller-manager/cmd/agent/main.go
```diff
@@ -118,25 +146,65 @@ func main() {
...
+	if err != nil {
 		log.Fatal(err)
 	}
 
-	log.Println("Discovery complete")
+	switch metalv1alpha1.State(*state) {
+	case metalv1alpha1.StateUnknown:
+		log.Printf("Creating resource via %q", *endpoint)
+
+		if err = create(*endpoint, s); err != nil {
```
smira, line 170: my only comment: should we create server in `decomissioned` state so that it goes through wipe before it is available?

smira (approved):
:+1:

app/metal-controller-manager/cmd/agent/main.go
andrewrynhard: Sounds like it could be a good idea. @rsmitty Thoughts?

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
```diff
@@ -363,6 +363,7 @@ func (r *MetalMachineReconciler) patchProviderID(ctx context.Context, cluster *c
 // patchServerInUse updates a server to mark it as "in use".
 func (r *MetalMachineReconciler) patchServerInUse(ctx context.Context, serverClass *metalv1alpha1.ServerClass, serverObj *metalv1alpha1.Server) error {
 	serverObj.Status.InUse = true
+	serverObj.Status.Clean = false
```
andrewrynhard, line 366: I'm not sure this is the best place for this. We don't really have a great way of knowing that an installed has taken place. That is why I am kind of torn on having this status. Maybe I am missing something still?

app/metal-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -53,6 +53,14 @@ func PartialEqual(a, b interface{}) bool {
 	old := reflect.ValueOf(a)
 	new := reflect.ValueOf(b)
 
+	if old.Kind() == reflect.Ptr {
+		old = old.Elem()
+	}
+
+	if new.Kind() == reflect.Ptr {
+		new = new.Elem()
+	}
```
andrewrynhard, line 62: Fixes a panic.

app/metal-controller-manager/cmd/agent/main.go
andrewrynhard: What if instead of marking it as decommissioned, which would require a second boot, we just run the cleanup code before we register the server?

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
smira: I think it's okay to over-wipe, but not okay to under-wipe.

We need to make sure though that we don't have nodes booted from the environment before server becomes `InUse`. So we should make sure iPXE doesn't deliver any image to the server (except for discovery one) until server is `InUse`.

app/metal-controller-manager/cmd/agent/main.go
smira: that sounds good as well!

Dockerfile
```diff
@@ -138,8 +138,8 @@ COPY --from=docker.io/autonomy/fhs:v0.2.0 / /
 COPY --from=assets /undionly.kpxe /var/lib/sidero/tftp/undionly.kpxe
 COPY --from=assets /undionly.kpxe /var/lib/sidero/tftp/undionly.kpxe.0
 COPY --from=assets /ipxe.efi /var/lib/sidero/tftp/ipxe.efi
-COPY --from=initramfs /initramfs.xz /var/lib/sidero/env/discovery/initramfs.xz
-ADD https://github.com/talos-systems/talos/releases/download/v0.4.1/vmlinuz /var/lib/sidero/env/discovery/vmlinuz
+COPY --from=initramfs /initramfs.xz /var/lib/sidero/env/agent/initramfs.xz
+ADD https://github.com/talos-systems/talos/releases/download/v0.4.1/vmlinuz /var/lib/sidero/env/agent/vmlinuz
```
andrewrynhard, line 142: It no longer makes sense to call this "discovery" since discovery isn't the only thing that this image does now. A more generic name would be better. This is a backwards compatible change since the agent and sidero are shipped as a single unit. In fact we can completely change the agent API at any time if we want.

Dockerfile
```diff
@@ -39,17 +45,11 @@ COPY --from=manifests-build /src/app/cluster-api-provider-sidero/config ./app/cl
...
-    && go get github.com/golang/protobuf/protoc-gen-go@v1.3
-COPY ./app/metal-controller-manager/internal/api/discovery.proto \
-      /src/app/metal-controller-manager/internal/api/discovery.proto
+COPY ./app/metal-controller-manager/internal/api/api.proto \
+  /src/app/metal-controller-manager/internal/api/api.proto
 RUN protoc -I/src/app/metal-controller-manager/internal/api \
-    --go_out=plugins=grpc,paths=source_relative:/src/app/metal-controller-manager/internal/api \
-    discovery.proto
+  --go_out=plugins=grpc,paths=source_relative:/src/app/metal-controller-manager/internal/api \
+  api.proto
```
andrewrynhard, line 52: See comment below.

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
```diff
@@ -214,7 +214,7 @@ func (r *MetalMachineReconciler) reconcileDelete(ctx context.Context, metalMachi
 			return ctrl.Result{}, err
 		}
 
-		serverResource.Status.InUse = false
+		serverResource.Status.NeedsWipe = true
```
andrewrynhard, line 217: We keep the server marked as in use as a way to prevent it from being mistakenly allocated again. We should only mark it as not in use once the reset process is done.

app/metal-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -100,13 +100,16 @@ type ServerSpec struct {
 
 // ServerStatus defines the observed state of Server.
 type ServerStatus struct {
-	Ready bool `json:"ready,omitempty"`
-	InUse bool `json:"inUse,omitempty"`
+	Ready     bool `json:"ready"`
+	InUse     bool `json:"inUse"`
+	NeedsWipe bool `json:"needsWipe"`
```
andrewrynhard, line 105: I want to propose `NeedsReset`. Reset sounds better than wipe IMHO. Not a huge deal, I can live with this if we think it is more accurate.

app/metal-controller-manager/cmd/agent/main.go
andrewrynhard: I decided not to wipe on registration since I am afraid of mistakenly wiping servers that contain important data, and are mistakenly booted into this environment (maybe a misconfigured DHCP option?).

app/metal-controller-manager/internal/api/api.proto
```diff
@@ -4,8 +4,10 @@ package api;
 
 option go_package = "github.com/talos-systems/sidero/app/metal-controller-manager/internal/api";
 
-service Discovery {
+service Agent {
   rpc CreateServer(CreateServerRequest) returns (CreateServerResponse);
+  rpc SetServerAsAvailable(SetServerAsAvailableRequest)
```
andrewrynhard, line 9: I _hate_ this RPC name. Any better ideas?

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
```diff
@@ -215,6 +215,7 @@ func (r *MetalMachineReconciler) reconcileDelete(ctx context.Context, metalMachi
 		}
 
 		serverResource.Status.InUse = false
```
andrewrynhard, line 217: I'm torn on leaving this here. Technically, the server is no longer in use in the sense that it is part of a cluster. On the other hand we run the risk of accidentally reusing the server before we have a chance to reset it. The alternative is that we remove this line, and mark the server as not in use when the agent updates the `NeedsWipe` (after it resets the node). Thoughts?

app/metal-controller-manager/api/v1alpha1/server_types.go
smira: Not a big deal, but probably having it inverse - like IsClean might work better as default value is more safe

smira (commented):
GitHub mobile doesn't let me submit review with empty comment

app/metal-controller-manager/cmd/agent/main.go
```diff
@@ -118,25 +146,64 @@ func main() {
...
+		}
 
-	if err = create(*endpoint, s); err != nil {
-		log.Fatal(err)
+		log.Println("Registration complete")
 	}
 
-	log.Println("Discovery complete")
+	wipe = procfs.ProcCmdline().Get(constants.AgentResetArg).First()
+	if wipe != nil {
```
Ulexus, line 178: For potentially-destructive things, I generally prefer an accept-list rather than a reject list.  (Is that the appropriate substitution for black-list, white-list?)

app/metal-controller-manager/internal/api/api.proto
Ulexus: `SetServerAvailability`?  I'm not sure I have a strong concept of its scope, as noted above.

app/metal-controller-manager/internal/server/server.go
```diff
@@ -84,7 +86,46 @@ func (s *server) CreateServer(ctx context.Context, in *api.CreateServerRequest)
...
+		return nil, err
+	}
+
+	obj.Status.NeedsWipe = false
+
+	if err := patchHelper.Patch(ctx, obj); err != nil {
+		return nil, err
+	}
+
+	log.Printf("Marked %q as clean", in.GetUuid())
```
Ulexus, line 124: Perhaps "NeedsWipe resolved" or "NeedsWipe removed" or "Removing NeedsWipe after performing wipe"

Ulexus (commented):
Feel free to ignore any of these based on my ignorance of things

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
smira: Should we search for !InUser and !NeedsWipe servers then?

app/metal-controller-manager/api/v1alpha1/server_types.go
smira: My point was not the name, but rather default value being "dirty" or "needs wipe" vs. default value being "clean". This is more safe.

app/metal-controller-manager/cmd/agent/main.go
smira: I think there should be no risk on wiping if we implement it correctly with `Server` lookup: if `Server` object doesn't exist (no such UUID), or `Server` is marked for wiping, wipe it.

If Server is `InUse` and doesn't need wiping - don't wipe it ;)

app/metal-controller-manager/internal/api/api.proto
smira: Should it be "MarkAsWiped"/"MarkAsClean" (as this is what it does?)

app/metal-controller-manager/internal/ipxe/ipxe_server.go
```diff
@@ -224,33 +184,121 @@ func parseMAC(s string) (net.HardwareAddr, error) {
...
+		if err != nil {
+			return nil, err
+		}
+	case server.OwnerReferences != nil:
+		env, err = newEnvironmentFromServerClass(c, server)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		env, err = newDefaultEnvironment(c)
```
smira, line 215: if we check here for `server.InUse` we can return 404 on attempt to boot server which is not use, this should resolve #53

app/metal-controller-manager/cmd/agent/main.go
andrewrynhard: I am less worried about things intended to be in the domain of Sidero, and more worried about machines that have accidentally been PXE booted with the Sidero services.

Ulexus: Yeah, it would be a pretty high price to pay for an easy-to-have-happen accident.  It seems like the simple existence check for the Server should be sufficient, though.  That's essentially an implicit "admission" flag.

app/metal-controller-manager/api/v1alpha1/server_types.go
Ulexus: Yeah, I'm just indicating that in this case, it may be worth that non-default valuation.

app/metal-controller-manager/internal/api/api.proto
andrewrynhard: `MarkAsClean`

app/metal-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -100,13 +100,16 @@ type ServerSpec struct {
 
 // ServerStatus defines the observed state of Server.
 type ServerStatus struct {
-	Ready bool `json:"ready,omitempty"`
-	InUse bool `json:"inUse,omitempty"`
+	Ready   bool `json:"ready"`
+	InUse   bool `json:"inUse"`
+	IsClean bool `json:"isClean"`
```
smira, line 105: :heart:

app/metal-controller-manager/internal/ipxe/ipxe_server.go
```diff
@@ -224,33 +192,118 @@ func parseMAC(s string) (net.HardwareAddr, error) {
...
+func newEnvironment(c client.Client, server *metalv1alpha1.Server) (env *metalv1alpha1.Environment, err error) {
+	// NB: The order of this switch statement is important. It defines the
+	// precedence of which environment to boot. The precedence is:
+	// - boot into the agent
+	// - skip if in use
+	// - boot into the environment defined by the server
+	// - boot into the environment defined by the server class
+	// - boot the default environment
+	switch {
+	case server == nil:
```
smira, line 206: check for `!IsClean` ?

app/metal-controller-manager/internal/server/server.go
```diff
@@ -43,48 +43,86 @@ func (s *server) CreateServer(ctx context.Context, in *api.CreateServerRequest)
...
+						SKUNumber:    in.GetSystemInformation().GetSkuNumber(),
+						Family:       in.GetSystemInformation().GetFamily(),
+					},
+					CPU: &metal1alpha1.CPUInformation{
+						Manufacturer: in.GetCpu().GetManufacturer(),
+						Version:      in.GetCpu().GetVersion(),
+					},
+				},
+				Status: metal1alpha1.ServerStatus{
+					IsClean: true,
```
smira, line 73: might need some `//TODO: ` here to change that once we do `Accepted:`

smira (commented):
I might have missed that, but I don't see an updated check for `IsClean` when picking up new servers, and I would probably set `IsClean = false` once server moved to `InUse` state

app/metal-controller-manager/internal/server/server.go
andrewrynhard: Some linter doesn't allow this, I will make a ticket to track.

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
```diff
@@ -276,11 +276,14 @@ func (r *MetalMachineReconciler) fetchServerFromClass(ctx context.Context, class
...
 		}
 
-		// double check that server hasn't been marked in use
 		if serverObj.Status.InUse {
 			continue
 		}
 
+		if !serverObj.Status.IsClean {
+			continue
+		}
```
andrewrynhard, line 285: @smira I think this is what you were looking for... Just added it.

smira (approved):

PR #66, opened 2020-09-17
chore: add additional logging for metadata server

### LINKED ISSUE #59: Improve Metadata Server Logging
We should provide better clarity in the pod logs on where we're actually failing when delivering metadata. Currently it only looks like: 
```
2020/09/11 14:20:06 received metadata request for uuid: 36353536-3135-5a43-3333-343956564a4e
```

rsmitty:
This PR adds a bunch of additional logging for the metadata server. This
is an attempt to make it clearer when we can't lookup a server or
bootstrap data for any given reason.

Will close #59 

### DIFF
--- a/app/metal-metadata-server/main.go
+++ b/app/metal-metadata-server/main.go
@@ -37,6 +37,11 @@ const (
 	metalVersion = "v1alpha1"
 )
 
+func throwError(w http.ResponseWriter, code int, err error) {
+	http.Error(w, err.Error(), code)
+	log.Println(err)
+}
+
 func main() {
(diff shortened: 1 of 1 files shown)


andrewrynhard (commented):
I feel like we could DRY this up a little more. Maybe a helper function that printed an error and reported an error?

```
func example(w http.ResponseWriter, code int, err error) {
    log.Println(err)

}
```

smira:
+1 to what Andrew said, we could also wrap this into a function which returns and error and set http status based on error plus log the error, looks like we need specific error for 404, and everything else is just 500?

smira (approved):

PR #67, opened 2020-09-17
feat: allow qualifers to be partially equal

andrewrynhard:
Instead of requiring a complete match of qualifiers, we now support the ability
to do partial matching. This means users can omit fiels in the CPU and System
Information qualifiers, and it will be a valid qualifier. Prior to this, every
field had to match.

### DIFF
--- a/app/metal-controller-manager/api/v1alpha1/server_types.go
+++ b/app/metal-controller-manager/api/v1alpha1/server_types.go
@@ -5,6 +5,8 @@
 package v1alpha1
 
 import (
+	"reflect"
+
 	corev1 "k8s.io/api/core/v1"
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


app/metal-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -34,11 +36,41 @@ type SystemInformation struct {
...
+func partialEqual(a, b interface{}) bool {
+	old := reflect.ValueOf(a)
+	new := reflect.ValueOf(b)
+
+	for i := 0; i < old.NumField(); i++ {
+		if old.Field(i).IsZero() {
+			// Skip zero values, since that indicates that the user did not supply
+			// the field, and does not want to compare it.
+			continue
+		}
```
andrewrynhard, line 61: There is one edge case here. If a user specifies a completely empty qualifier in YAML (e.g. `cpu: {}`), then this will still match. I would argue that that is user error, but we _could_ build in a safeguard and require that at least one field isn't the zero value.

app/metal-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -34,11 +36,41 @@ type SystemInformation struct {
...
 type CPUInformation struct {
 	Manufacturer string `json:"manufacturer,omitempty"`
 	Version      string `json:"version,omitempty"`
 }
 
+func (a *CPUInformation) PartialEqual(b *CPUInformation) bool {
+	return partialEqual(a, b)
+}
+
+func partialEqual(a, b interface{}) bool {
```
andrewrynhard, line 52: I am sure this function can be picked apart if we wanted to make it super generic. I could spend the time on that, or we could just use it knowing that it only works for structs with a flat list of string fields.

smira (approved):
:+1:

smira (approved):

PR #72, opened 2020-09-18
fix: ensure proper checking of "in-use" status when fetching metadata

rsmitty:
This PR fixes a bug where users could have orphaned metalmachines that still
had ownerrefs to a server that had alredy been made available for another
cluster. This led to incorrect metadata being returned. We now bail and
error early if we encounter multiple ownerrefs in the list pointing to
the same server.

### DIFF
--- a/app/metal-metadata-server/main.go
+++ b/app/metal-metadata-server/main.go
@@ -66,6 +66,8 @@ func FetchConfig(w http.ResponseWriter, r *http.Request) {
 				"received metadata request with empty uuid",
 			),
 		)
+
+		return
 	}
 
 	log.Printf("received metadata request for uuid: %s", uuid)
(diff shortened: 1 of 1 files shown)


app/metal-metadata-server/main.go
```diff
@@ -147,406 +153,431 @@ func FetchConfig(w http.ResponseWriter, r *http.Request) {
...
 
-				err = runtime.DefaultUnstructuredConverter.FromUnstructured(ownerItem.(map[string]interface{}), tempOwnerRef)
-				if err != nil {
-					throwError(
-						w,
-						http.StatusInternalServerError,
-						fmt.Errorf("failure converting ownerItem to ownerRef: %s", err),
-					)
+	// If we have a match, fetch the bootstrap data from machine resource that owns this metal machine
+	ownerList, present, err := unstructured.NestedSlice(matchingMetalMachine.Object, "metadata", "ownerReferences")
```
rsmitty, line 190: From here down looks so much worse that it is. It's really just an un-indentation from moving the for loop that goes through metal machines.

smira (approved):
I feel this needs to be broken down in many functions, diff is really hard to follow..

PR #78, opened 2020-09-22
chore: merge Sfyra into Sidero repository

### LINKED ISSUE #77: Move to Sfyra to Sidero

smira:
Mostly only file renaming and other non-functional changes.

Incorporated fix from https://github.com/talos-systems/sfyra/pull/21.

Fixes #77 


app/metal-controller-manager/internal/ipxe/ipxe_server.go
```diff
@@ -229,7 +229,9 @@ func newEnvironment(c client.Client, server *metalv1alpha1.Server) (env *metalv1
 		if err != nil {
 			return nil, err
 		}
-	default:
+	}
```
smira, line 232: this fixes a bug: `server.OwnerReferences` might be `!= nil`, but environment might not be available anyways, so we have to always fall back to the default environment.

andrewrynhard (approved):

PR #105, opened 2020-09-30
test: add scale test

andrewrynhard:
Adds a test for scaling workers and control plane.

### DIFF
--- a/app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
+++ b/app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
@@ -58,7 +58,7 @@ func (r *MetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, err
 
 	err = r.Get(ctx, req.NamespacedName, metalMachine)
 	if apierrors.IsNotFound(err) {
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: constants.DefaultRequeueAfter}, nil
 	}
 
 	if err != nil {
@@ -67,11 +67,15 @@ func (r *MetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, err
 
 	machine, err := util.GetOwnerMachine(ctx, r.Client, metalMachine.ObjectMeta)
 	if err != nil {
-		return ctrl.Result{}, err
+		r.Log.Error(err, "Failed to get machine")
+
+		return ctrl.Result{RequeueAfter: constants.DefaultRequeueAfter}, nil
 	}
 
 	if machine == nil {
-		return ctrl.Result{}, fmt.Errorf("no ownerref for machine %s", metalMachine.ObjectMeta.Name)
+		logger.Info("No ownerref for metalmachine")
+
+		return ctrl.Result{RequeueAfter: constants.DefaultRequeueAfter}, nil
 	}
 
 	logger = logger.WithName(fmt.Sprintf("machine=%s", machine.Name))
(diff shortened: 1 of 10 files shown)


app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
```diff
@@ -58,7 +59,7 @@ func (r *MetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, err
 
 	err = r.Get(ctx, req.NamespacedName, metalMachine)
 	if apierrors.IsNotFound(err) {
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: 20 * time.Second}, nil
```
Unix4ever, line 62: Does it worths it to move this requeue time to some constants?

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
```diff
@@ -67,11 +68,15 @@ func (r *MetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, err
...
 	if err != nil {
-		return ctrl.Result{}, err
+		r.Log.Error(err, "Failed to get machine")
+
+		return ctrl.Result{RequeueAfter: 20 * time.Second}, nil
 	}
 
 	if machine == nil {
-		return ctrl.Result{}, fmt.Errorf("no ownerref for machine %s", metalMachine.ObjectMeta.Name)
+		r.Log.Info("No ownerref for metalmachine", "metalmachine", metalMachine.ObjectMeta.Name)
```
Unix4ever, line 77: Logger is already initialized as:
```go
logger := r.Log.WithValues("metalmachine", req.NamespacedName)
```
Does it override this variable? Are these variable values different?
If they're different maybe it makes sense to initialize logger with `metalMachine.ObjectMeta.Name` otherwise?

Unix4ever (approved):
Just made similar changes at the same place, but was too slow :)

One more thing which I also wasn't sure about. If we should have exponential backoff for machine not being accepted here:
```go
if !serverResource.Spec.Accepted {
			return ctrl.Result{}, fmt.Errorf("specified serverref is a non-accepted server")				return ctrl.Result{}, fmt.Errorf("specified serverref is a non-accepted server")
```
Just trying to see that from UX perspective. I guess it would be nice for user if Sidero picks the machine as fast as possible, but at the same it may be too bad to spin in a cycle trying to add machine which may never be accepted by user.

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
andrewrynhard: Makes perfect sense. Thanks!

andrewrynhard:
> ```go
> return ctrl.Result{}, fmt.Errorf("specified serverref is a non-accepted server")
> ```

This branch tripped me up as well when I first saw it. @rsmitty might have a better opinion on it. There are tradeoffs either way I think. If the user never does update it, then how often should we reconcile? I guess its not horrible to retry every 20 seconds. I don't know what the right thing is here TBH :slightly_frowning_face:.

Unix4ever:
Prob it makes sense to get an idea on what is actual limit for exponential backoff. If it's 1 hour for example, I guess it's completely unacceptable and we should use requeue there.

andrewrynhard:
> Prob it makes sense to get an idea on what is actual limit for exponential backoff. If it's 1 hour for example, I guess it's completely unacceptable and we should use requeue there.

Agree.

PR #108, opened 2020-09-30
chore: refactor metadata server code

### LINKED ISSUE #74: [metadata server] break main apart into functions
The metadata server code is nigh impossible to read. It should really be broken up to be a bit more modular and readable.

rsmitty:
This breaks apart the metadata server to be more modular

Will close #74


andrewrynhard:
Haven't looked at the code yet, but `refactor` is a conventional commit type. This should be `refactor: ...`

smira (approved):
I like the way it looks now much better than what it was before :+1: 

Few thoughts (not blocking the PR):

We might want to separate HTTP handling from the actual business logic. We can do HTTP handler which parses request, calls function which returns config and error back, and acts on the error (error might contain status code -> print status code, otherwise just 500 with error text). This allows the main method to be testable outside of the HTTP handler context, and probably easier to follow, as the function just returns an error (without multiple `throwError` all over the code)

rsmitty:
@smira good thought, I'll drop a ticket for that separation as a new task.

PR #113, opened 2020-10-02
feat: power down servers !InUse && IsClean

smira:
Basically this repeats the power off action which should be done by the
agent image. In case of Sfyra and QEMU VMs, they can't be shutdown via
ACPI, so this code is going to bring Sfyra case closer to the bare metal
case, plus it should handle power off if Server is registered manually
(without agent).

Also mark servers which are `InUse && !IsClean` as `Ready` to assert
that all the servers are ready in the test.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -189,7 +189,7 @@ COPY --from=fmt-build /src /
 # The markdownlint target performs linting on Markdown files.
 #
 FROM node:8.16.1-alpine AS lint-markdown
-RUN npm install -g markdownlint-cli
+RUN npm install -g markdownlint-cli@0.23.2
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -189,7 +189,7 @@ COPY --from=fmt-build /src /
 # The markdownlint target performs linting on Markdown files.
 #
 FROM node:8.16.1-alpine AS lint-markdown
-RUN npm install -g markdownlint-cli
+RUN npm install -g markdownlint-cli@0.23.2
```
smira, line 192: this fixes the lint-markdown step

andrewrynhard (approved):

PR #128, opened 2020-10-07
test: break serverclass creation into function and allow dummy servers

rsmitty:
This PR moves the serverClass creation into a function that can be
called by multiple tests. Additionally, I added in a function to create
dummy servers that aren't backed by QEMU VMs. Helpful as we develop the
configPatching tests.

### DIFF
--- a/sfyra/pkg/tests/server.go
+++ b/sfyra/pkg/tests/server.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/talos-systems/sidero/app/metal-controller-manager/api/v1alpha1"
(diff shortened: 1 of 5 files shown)


sfyra/pkg/tests/server.go
```diff
@@ -196,3 +197,24 @@ func TestServersReady(ctx context.Context, metalClient client.Client) TestFunc {
 		}))
 	}
 }
+
+// createDummyServers will submit servers with dummy info that are not tied to QEMU VMs.
+// A number of these, based on "count" will be created.
+// These can be targeted by the spec passed in or the label "dummy-server".
+
+//nolint: deadcode,unused
```
rsmitty, line 205: Ignored these linters temporarily. We'll be using this function very soon.

smira (approved):

PR #131, opened 2020-10-08
feat: support server acceptance

### LINKED ISSUE #122: Add accepted field to server spec
Accepted means that we are free to clean the node any time that we would like. Alternatives for "accepted" might be "approved" or "admitted". We should decide on a name since naming is hard and changing it would be breaking. This new field should be part of the server spec so that users can control it.

rsmitty:
This PR adds a bool field to the server spec of `accepted`. This allows
users to adopt a server that has been registered into the sidero
environment. Doing so allows for a server's full lifecycle to be
managed. Users can either edit a server resource to update the accepted
field, or create the server resource manually and supply that with their
spec.

The PR also adds some edge case handling around the sidero API server to
check to make sure a server is accepted before allowing a wipe directive
to be returned, as well as a check for acceptance when trying to use a
serverref in the sidero capi provider.

Will close #122

### DIFF
--- a/sfyra/pkg/tests/server.go
+++ b/sfyra/pkg/tests/server.go
@@ -21,6 +21,7 @@ import (
 	talosconfig "github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1"
 	"gopkg.in/yaml.v3"
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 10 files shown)


sfyra/pkg/tests/server_class.go
```diff
@@ -41,6 +42,23 @@ func TestServerClassDefault(ctx context.Context, metalClient client.Client, vmSe
 		serverClass, err := createServerClass(ctx, metalClient, serverClassName, classSpec)
 		require.NoError(t, err)
 
+		// verify sc is initially empty
+		assert.Len(t, append(serverClass.Status.ServersAvailable, serverClass.Status.ServersInUse...), 0)
```
andrewrynhard, line 46: :+1:

andrewrynhard: Only thing I might suggest is that we do something similar to this but in the server test. It will kick off the server reset sooner. I don't think it matters too much in the end.

andrewrynhard (approved):

sfyra/pkg/tests/server.go
```diff
@@ -177,6 +177,51 @@ func TestServerPatch(ctx context.Context, metalClient client.Client, talosInstal
 	}
 }
 
+// TestServerAcceptance makes sure the accepted bool works.
+func TestServerAcceptance(ctx context.Context, metalClient client.Client, vmSet *vm.Set) TestFunc {
+	return func(t *testing.T) {
+		numNodes := len(vmSet.Nodes())
+
+		servers := &v1alpha1.ServerList{}
```
smira, line 185: the only problem here is that this test is not idempotent if we re-run it. 

I don't have any good idea to keep it both idempotent and verify that accepted is set to false... except if we just create a mock server and verify accepted is false?

sfyra/pkg/tests/server_class.go
smira: this condition will fail if we re-run the tests (server class will already grab some servers)

my only suggestion is to create and destroy set of "fake" servers specifically to test this feature

andrewrynhard: That makes sense.

andrewrynhard (approved):

smira (approved):

PR #134, opened 2020-10-08
chore: refactor `reconcile` method in `environment_controller.go`

Unix4ever:
This should make method shorter and (hopefully) more readable.
Tried to elliminate duplicated blocks.

### DIFF
--- a/app/metal-controller-manager/controllers/environment_controller.go
+++ b/app/metal-controller-manager/controllers/environment_controller.go
@@ -16,7 +16,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"github.com/hashicorp/go-multierror"
+	multierror "github.com/hashicorp/go-multierror"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


Unix4ever:
Or I can split that to separate methods, what do you think?

app/metal-controller-manager/controllers/environment_controller.go
```diff
@@ -93,43 +94,42 @@ func (r *EnvironmentReconciler) reconcile(req ctrl.Request) (ctrl.Result, error)
...
+			mu.Lock()
+			conditions = append(conditions, condition)
+			mu.Unlock()
+		}
 
+		saveAsset := func() {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				if err := save(ctx, assetTask.Asset, file); err != nil {
```
smira, line 120: this is going to be a problem - file is outer variable which gets changes in the for loop and used in goroutine, so it will change the value while goroutine is running

Unix4ever: ah, right. I've missed that wg is being waited after the loop.

Unix4ever: fixed

andrewrynhard (approved):

PR #141, opened 2020-10-09
hack: try with new talosctl

smira:
Hack hack hack


smira:
binary from https://github.com/talos-systems/talos/pull/2622

PR #143, opened 2020-10-09
feat: support config patches at the serverclass level

rsmitty:
This PR allows configPatches to be specified for serverclasses


app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
```diff
@@ -240,6 +240,13 @@ func (r *MetalMachineReconciler) reconcileDelete(ctx context.Context, metalMachi
 		}
 
 		if err := r.Get(ctx, namespacedName, serverResource); err != nil {
+			// bail early if the server can't be fetch. this likely means we've deleted the server from underneath already.
+			if apierrors.IsNotFound(err) {
```
rsmitty, line 244: This just returns early to handle the case where we were trying to patch a server after it had already been cleaned up.

andrewrynhard (approved):

PR #153, opened 2020-10-12
fix: handle agent errors gracefully

andrewrynhard:
Instead of exiting (which causes a kernel panic), we shutdown instead.

### DIFF
--- a/app/metal-controller-manager/cmd/agent/main.go
+++ b/app/metal-controller-manager/cmd/agent/main.go
@@ -178,28 +178,48 @@ func reconcileIPs(endpoint string, s *smbios.Smbios, ips []net.IP) error {
 	return nil
 }
 
+func shutdown(err error) {
+	if err != nil {
+		log.Println(err)
+
+		for i := 10; i >= 0; i-- {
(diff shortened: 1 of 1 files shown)


app/metal-controller-manager/cmd/agent/main.go
```diff
@@ -178,57 +178,75 @@ func reconcileIPs(endpoint string, s *smbios.Smbios, ips []net.IP) error {
 	return nil
 }
 
+func shutdown(err error) {
+	if err != nil {
```
smira, line 182: I guess we should print the error here?

smira:
not related to this PR, but one little change I wish we could do is to extract grpc client creation to the common place, so that every function doesn't have to do that.

smira (approved):

smira:
one more idea is to wrap the `main` into a function which returns an error and do `shutdown(f())`

andrewrynhard:
> one more idea is to wrap the `main` into a function which returns an error and do `shutdown(f())`

Sorry, I missed this. The code could use some cleanup.

PR #157, opened 2020-10-13
test: check if servers are reset on acceptance

andrewrynhard:
When a server is dirty and not accepted, it should be reset after being accepted.

### DIFF
--- a/sfyra/pkg/tests/reset.go
+++ b/sfyra/pkg/tests/reset.go
@@ -48,7 +48,7 @@ func TestServerReset(ctx context.Context, metalClient client.Client, vmSet *vm.S
 			require.NoError(t, err)
 		}
 
-		err = retry.Constant(10*time.Minute, retry.WithUnits(10*time.Second)).Retry(func() error {
+		err = retry.Constant(5*time.Minute, retry.WithUnits(10*time.Second)).Retry(func() error {
 			var servers sidero.ServerList
 
 			err = metalClient.List(ctx, &servers)
--- a/sfyra/pkg/tests/server.go
+++ b/sfyra/pkg/tests/server.go
@@ -260,6 +260,56 @@ func TestServerAcceptance(ctx context.Context, metalClient client.Client, vmSet 
 	}
 }
 
+// TestServerResetOnAcceptance tests that servers are reset when accepted.
+func TestServerResetOnAcceptance(ctx context.Context, metalClient client.Client) TestFunc {
+	return func(t *testing.T) {
+		serverList := &v1alpha1.ServerList{}
+
+		err := metalClient.List(ctx, serverList)
+		require.NoError(t, err)
+
+		servers := []v1alpha1.Server{}
+
+		for _, server := range serverList.Items {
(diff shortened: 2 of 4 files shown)


sfyra/pkg/tests/server.go
```diff
@@ -260,6 +260,41 @@ func TestServerAcceptance(ctx context.Context, metalClient client.Client, vmSet
 	}
 }
 
+// TestServerResetOnAcceptance tests that servers are reset when accepted.
+func TestServerResetOnAcceptance(ctx context.Context, metalClient client.Client, vmSet *vm.Set) TestFunc {
```
smira, line 264: `vmSet` seems to be unused

sfyra/pkg/tests/server.go
```diff
@@ -260,6 +260,41 @@ func TestServerAcceptance(ctx context.Context, metalClient client.Client, vmSet
...
+		err := metalClient.List(ctx, servers)
+		require.NoError(t, err)
+
+		for _, server := range servers.Items {
+			server := server
+
+			patchHelper, err := patch.NewHelper(&server, metalClient)
+			require.NoError(t, err)
+
+			server.Spec.Accepted = true
```
smira, line 276: we should count number of servers transitioned into `Accepted` here, and in the next `retry` block test only those servers to get things idempotent

andrewrynhard: I'm not sure I understand fully. I will try and you can tell me if I got it right.

sfyra/pkg/tests/reset.go
```diff
@@ -48,7 +48,7 @@ func TestServerReset(ctx context.Context, metalClient client.Client, vmSet *vm.S
 			require.NoError(t, err)
 		}
 
-		err = retry.Constant(10*time.Minute, retry.WithUnits(10*time.Second)).Retry(func() error {
+		err = retry.Constant(5*time.Minute, retry.WithUnits(10*time.Second)).Retry(func() error {
```
andrewrynhard, line 51: There is no reason that a reset of a single server should take more than a minute let alone five. This will just keep CI from hanging.

sfyra/pkg/tests/server.go
```diff
@@ -260,6 +260,56 @@ func TestServerAcceptance(ctx context.Context, metalClient client.Client, vmSet
...
+
+				if err := metalClient.Get(ctx, types.NamespacedName{Name: servers[i].Name, Namespace: servers[i].Namespace}, &s); err != nil {
+					return retry.UnexpectedError(err)
+				}
+
+				if !s.Status.IsClean {
+					return retry.ExpectedError(fmt.Errorf("server %q is not clean", s.Name))
+				}
+
+				servers = append(servers[:i], servers[i+1:]...)
```
smira, line 305: I think this skips over the servers as we remove them (as it skips over the next server being removed), but I don't think we need to remove from the list really

rsmitty (approved):

smira (approved):

PR #173, opened 2020-10-21
chore: add support for control plane port

rsmitty:
We were missing support for tweaking the load balancer port


smira (approved):

andrewrynhard (dismissed):
LGTM, but would you mind amending and making this a feat?

smira (approved):

PR #183, opened 2020-10-29
fix: log error with error log

Ulexus:
Log actual error with error log.

### DIFF
--- a/app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
+++ b/app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
@@ -220,7 +220,7 @@ func (r *MetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, err
 
 	err = r.patchProviderID(ctx, cluster, metalMachine)
 	if err != nil {
-		logger.Info("Failed to set provider ID")
+		logger.Info("Failed to set provider ID", "error", err)
 
 		return ctrl.Result{RequeueAfter: constants.DefaultRequeueAfter}, nil
 	}


andrewrynhard (approved):

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
```diff
@@ -220,7 +220,7 @@ func (r *MetalMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, err
 
 	err = r.patchProviderID(ctx, cluster, metalMachine)
 	if err != nil {
-		logger.Info("Failed to set provider ID")
+		logger.Info("Failed to set provider ID:", err)
```
andrewrynhard, line 223: I think this will panic at runtime. This library requires key value pairs after the first arg... so something like:

```
		logger.Info("Failed to set provider ID", "error", err)
```

andrewrynhard (changes requested):

app/cluster-api-provider-sidero/controllers/metalmachine_controller.go
andrewrynhard: FWIW ^ is the convention (I believe) we have used.

Ulexus: It definitely doesn't panic, as I used this to figure out what what going on, but I'm happy to conform.

andrewrynhard: That is interesting. I believe you, just breaks my understanding :smile: ....

FWIW, here is the interface definition
```
	// The msg argument should be used to add some constant description to
	// the log line.  The key/value pairs can then be used to add additional
	// variable information.  The key/value pairs should alternate string
	// keys and arbitrary values.
	Info(msg string, keysAndValues ...interface{})
```

:man_shrugging: I am happy either way.

andrewrynhard (approved):

PR #184, opened 2020-10-29
docs: add non-UEFI clause to PXE example

Ulexus:
Revise the example DHCP config to handle BIOS-based PXE clients.


andrewrynhard (approved):
:+1:

PR #186, opened 2020-10-29
docs: expand server acceptance doc

### LINKED ISSUE #177: Update bootstrapping guide for acceptance
https://www.sidero.dev/docs/v0.1/configuration/servers/#server-acceptance talks about the criticality of server acceptance.  (it would be clearer if it said "accepted: must be set to true" on that page - I thought acceptance was a process that needed to occur, not a config.)

But the bootstrapping guide https://www.sidero.dev/docs/v0.1/guides/bootstrapping/ doesn't mention acceptance.
It talks about registering servers - which led me to wonder are all registered servers accepted?

The answer is You can set it up to automatically accept any new server, but the process should generally be:  machine boots, gets registered into Sidero.  Then, some business process determines whether that server should be "accepted" as a valid server into Sidero.  A registered server is "known" to Sidero (registered with machine info), but it will not be manipulated in any way until it is "accepted"

So the bootstrapping guide needs to cover accepting the servers.

So in this Issue:
- update /configuration/servers to say that accepted:true must be set
- update /guides/bootstrapping/ to detail how to accept servers, and the concept of doing so
- update Concepts page to define "accepted", "allocated", "clean" states.

What is effect if an "Accepted" server is later modified to "Accepted:false"? Is it de-provisioned? Wiped? Or just frozen in its current state? How about if its currently allocated? DOes that change it? Document that in concepts too, I think.

Ulexus:
Expand the Server acceptance documentation to better describe the
intention and function.

This addresses but does not completely resolve #177.  Additional
lifecycle documentation (which does not seem to appropriate here)
should be added to describe how to take a `Server` out of commission.

### DIFF
--- a/docs/website/content/docs/v0.1/Configuration/servers.md
+++ b/docs/website/content/docs/v0.1/Configuration/servers.md
@@ -39,4 +39,20 @@ spec:
 ## Server Acceptance
 
 In order for a server to be eligible for consideration, it _must_ be `accepted`.
+This is an important separation point which all `Server`s must pass.
+Before a `Server` is accepted, no write action will be performed against it.
+Thus, it is safe for a computer to be added to a network on which Sidero is operating.
+Sidero will never write to or wipe any disk on a computer which is not marked as `accepted`.
+
+This can be tedious for systems in which all attached computers should be considered to be under the control of Sidero.
+Thus, you may also choose to automatically accept any machine into Sidero on its discovery.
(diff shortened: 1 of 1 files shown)


andrewrynhard (approved):
:+1: This makes we wonder if we should implement a validation webhook and prevent changing of accepted if it is in use.

rsmitty (approved):

Ulexus:
Yeah, I agree:  an externally-fed webhook would be ideal for a standard use-case of acceptance.

PR #209, opened 2020-11-10
feat: make "insecure-wipe" default, use new fast wipe method

### LINKED ISSUE #207: Improve reset/wipe.

smira:
Use new feature of go-blockdevice to implement the "fast wipe" in Sidero
and make it default one: https://github.com/talos-systems/go-blockdevice/pull/13

This new fast wipe is as good as zeroing out with most modern SSDs, but
not as secure, so we can call it "insecure wipe", but make it default to
improve the customer experience.

Fixes #207

### DIFF
--- a/app/metal-controller-manager/cmd/agent/main.go
+++ b/app/metal-controller-manager/cmd/agent/main.go
@@ -5,9 +5,7 @@
 package main
 
 import (
-	"bytes"
 	"context"
-	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


andrewrynhard (approved):

andrewrynhard (approved):
I'm a little hesitant to make insecure the default, but I think at this point it is just too slow to do it by default. LGTM.

PR #211, opened 2020-11-10
feat: boot servers via PXE only once by default

### LINKED ISSUE #205: Sidero: PXE-boot "once" into the environment

smira:
Fixes #205

This change makes sure that server gets booted into the environment only
once (unless `.Spec.PXEBootAlways` is set).

If IPMI is used, there're no functional changes, as servers are supposed
to be configured to boot from disk first, and Sidero forces boot from
network as needed.

Without IPMI, now it's possible to set server to boot from network all
the time, and once provisioned, server will boot from disk.

### DIFF
--- a/app/metal-controller-manager/api/v1alpha1/server_types.go
+++ b/app/metal-controller-manager/api/v1alpha1/server_types.go
@@ -89,19 +89,38 @@ type ServerSpec struct {
 	ManagementAPI     *ManagementAPI          `json:"managementApi,omitempty"`
 	ConfigPatches     []ConfigPatches         `json:"configPatches,omitempty"`
 	Accepted          bool                    `json:"accepted"`
+	PXEBootAlways     bool                    `json:"pxeBootAlways,omitempty"`
 }
 
-// ConditionPowerCycle is used to control the powercycle flow.
-const ConditionPowerCycle clusterv1.ConditionType = "PowerCycle"
+const (
+	// ConditionPowerCycle is used to control the powercycle flow.
+	ConditionPowerCycle clusterv1.ConditionType = "PowerCycle"
+	// ConditionPXEBooted is used to record the fact that server got PXE booted.
+	ConditionPXEBooted clusterv1.ConditionType = "PXEBooted"
+)
 
 // ServerStatus defines the observed state of Server.
 type ServerStatus struct {
-	Ready      bool                  `json:"ready"`
-	InUse      bool                  `json:"inUse"`
-	IsClean    bool                  `json:"isClean"`
+	// Ready is true when server is accepted and in use.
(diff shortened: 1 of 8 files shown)


app/metal-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -89,16 +89,25 @@ type ServerSpec struct {
...
+	// ConditionPXEBooted is used to record the fact that server got PXE booted.
+	ConditionPXEBooted clusterv1.ConditionType = "PXEBooted"
+)
 
 // ServerStatus defines the observed state of Server.
 type ServerStatus struct {
-	Ready      bool                  `json:"ready"`
-	InUse      bool                  `json:"inUse"`
-	IsClean    bool                  `json:"isClean"`
+	// +optional
```
smira, line 104: this is crucial to get correct `patch` operations working on the `Status`, as otherwise it fails with `.status.ready field being required`.

app/metal-controller-manager/controllers/server_controller.go
```diff
@@ -61,24 +68,26 @@ func (r *ServerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
-	f := func(ready, requeue bool) (ctrl.Result, error) {
-		s.Status.Power = "off"
+	s.Status.Power = "off"
 
-		poweredOn, err := mgmtClient.IsPoweredOn()
-		if err != nil {
-			s.Status.Power = "unknown"
-		}
+	poweredOn, powerErr := mgmtClient.IsPoweredOn()
```
smira, line 73: I understand that power state was read in `defer` to make sure we read the status after we do any changes, but looks like reconcile loop is anyway launched once more, so next interation reads latest state, but this change allows to read the power state only once for the reconcile loop

app/metal-controller-manager/controllers/server_controller.go
```diff
@@ -61,24 +68,26 @@ func (r *ServerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 		s.Status.Ready = ready
 
 		result := ctrl.Result{Requeue: requeue}
 
-		if err := r.Status().Update(ctx, &s); err != nil {
-			return result, err
+		if err := patchHelper.Patch(ctx, &s, patch.WithOwnedConditions{
+			Conditions: []clusterv1.ConditionType{metalv1alpha1.ConditionPowerCycle},
+		}); err != nil {
+			return result, errors.WithStack(err)
```
smira, line 90: `errors.WithStack` embeds original stack trace in the log message, not perfect, but better than nothing I guess

andrewrynhard (approved):

PR #230, opened 2020-11-24
test: add simulated power management failures in testing mode

smira:
This injects power management failures in Sfyra (under test). By default
failures are not injected.

Two class of failures are supported:

* explicit (when client returns an error)
* silent (when client doesn't do the job, but error is not returned)


rsmitty (dismissed):

andrewrynhard (approved):

smira:
I don't want to merge this PR just yet until we have a better stability of Sidero, as this might introduce some failures into Sfyra we would have to dig into to understand if it's introduced failure or what not. should be easy enough to merge it later

sfyra/cluster.yaml
```diff
@@ -0,0 +1,108 @@
+apiVersion: cluster.x-k8s.io/v1alpha3
```
smira, line 1: TODO: remove this

PR #249, opened 2020-12-10
feat: send heartbeat from agent while wipe is in progress

Unix4ever:
This allows to manage 'wipe timeout' better - no matter how long wiping
might take, we just have to wait for reboot timeout, which is 20 minutes by default, and
then if wipe 'heartbeat' never came, we can treat that as timeout and restart
the PXE reboot sequence.

Fixes: https://github.com/talos-systems/sidero/issues/235


app/metal-controller-manager/cmd/agent/main.go
```diff
@@ -247,7 +248,48 @@ func mainFunc() error {
...
+
+		go func() {
+			defer wg.Done()
+
+			for {
+				select {
+				case <-ticker.C:
+					callCtx, cancel := context.WithTimeout(ctx, heartbeatInterval)
+
+					if _, err := client.WipeHeartbeat(callCtx, &api.WipeHeartbeatRequest{Uuid: uuid.String()}); err != nil {
```
smira, line 277: I think we should send first heartbeat asap, without waiting for the first interval so that once agent boots up, it resets the timeout

(should be as easy as moving this code out of `select` to the `for` loop before `select`

app/metal-controller-manager/internal/api/api.proto
```diff
@@ -42,11 +43,14 @@ message Address {
 message CreateServerResponse {
   bool wipe = 1;
   bool insecure_wipe = 2;
+  double reboot_timeout = 3;
 }
 
 message MarkServerAsWipedRequest { string uuid = 1; }
+message WipeHeartbeatRequest { string uuid = 1; }
```
andrewrynhard, line 50: I think we should keep this generic and just call it `Heartbeat`. I can see the heartbeat being useful outside of wiping.

andrewrynhard (commented):
Looks great. This should be a `feat` IMO.

smira (approved):

PR #252, opened 2020-12-14
fix: overwrite kernel args from --extra-agent-kernel-args

### LINKED ISSUE #251: serial console setting does not work with --extra-agent-kernel-args
By default, agent environment kernel parameters has `console=ttyS0`.
With newly added `--extra-agent-kernel-args`, enabled the ability to add  `console=ttyS1`. (the config working in my servers)
The implemenation ended up with both parameters present, I'm not sure how kernel handles this, but in real test, I still could not get serial console output during agent running.

From from https://www.kernel.org/doc/html/v5.9/admin-guide/serial-console.html, below is mentioned:

    Note that you can only define one console per device type (serial, video).

smira:
This allows to override default options (e.g. `console=ttyS0`) with
provided values. Make sure that full value is given for each option, as
any default values for the option are dropped if specified in
`--extra-agent-kernel-args`, e.g.
`--extra-agent-kernel-args "console=tty0 console=ttyS1 panic=5"`.

See also talos-systems/go-procfs#5

Fixes #251


Unix4ever (approved):

Makefile
```diff
@@ -9,7 +9,7 @@ MODULE := $(shell head -1 go.mod | cut -d' ' -f2)
 
 ARTIFACTS := _out
 PKGS ?= ./...
-TALOS_RELEASE ?= v0.7.0
```
smira, line 12: this is side-effect of pulling in new go-procfs with `.Get()` fix - it also pulls in new default KSPP args which are not compatible with 0.7

I have an issue in our backlog to decouple default args from go-procfs

app/metal-controller-manager/internal/ipxe/ipxe_server.go
```diff
@@ -50,7 +52,7 @@ exit
 
 var (
 	apiEndpoint          string
-	extraAgentKernelArgs []string
```
smira, line 53: change format to whatever kernel args format is (with spaces), as commas have special meaning in kernel args

hack/scripts/integration-test.sh
```diff
@@ -18,9 +18,6 @@ function build_registry_mirrors {
 
       REGISTRY_MIRROR_FLAGS="${REGISTRY_MIRROR_FLAGS} --registry-mirrors ${registry}=http://${addr}:5000"
     done
-
-    local addr=`python3 -c "import socket; print(socket.gethostbyname('registry.ci.svc'))"`
```
smira, line 22: this is long gone, just gc'ed it

hack/scripts/integration-test.sh
```diff
@@ -38,5 +35,5 @@ ${PREFIX} "${INTEGRATION_TEST}" test integration \
     --talosctl-path "${TALOSCTL}" \
     --clusterctl-config "${CLUSTERCTL_CONFIG}" \
     --power-simulated-explicit-failure-prob=0.1 \
-    --power-simulated-silent-failure-prob=0.1 \
```
smira, line 41: brought it to zero as it introduces flakes and we don't need them

Ulexus (approved):

PR #264, opened 2020-12-28
fix: prevent server orphaning

Unix4ever:
Fixes: https://github.com/talos-systems/sidero/issues/155

Introduce server finalizers, which will block server deletion until
server is deallocated.

### DIFF
--- a/app/metal-controller-manager/controllers/server_controller.go
+++ b/app/metal-controller-manager/controllers/server_controller.go
@@ -25,6 +25,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
@@ -33,6 +34,10 @@ import (
 	metalv1alpha1 "github.com/talos-systems/sidero/app/metal-controller-manager/api/v1alpha1"
 	"github.com/talos-systems/sidero/app/metal-controller-manager/internal/power/metal"
 	"github.com/talos-systems/sidero/app/metal-controller-manager/pkg/constants"
+)
+
+const (
+	serverBindingFinalizer = "storage.finalizers.server.k8s.io"
 )
 
 // ServerReconciler reconciles a Server object.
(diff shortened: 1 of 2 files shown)


app/metal-controller-manager/controllers/server_controller.go
```diff
@@ -129,6 +145,21 @@ func (r *ServerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 	}
 
+	if s.ObjectMeta.DeletionTimestamp.IsZero() {
+		if finalizerIndex(s.ObjectMeta.Finalizers, serverBindingFinalizer) == -1 {
```
smira, line 149: There are helper methods for that already AddFinalizer, RemoveFinalizer

app/metal-controller-manager/controllers/server_controller.go
```diff
@@ -129,6 +145,21 @@ func (r *ServerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			s.ObjectMeta.Finalizers = append(s.ObjectMeta.Finalizers, serverBindingFinalizer)
+			if err := r.Update(ctx, &s); err != nil {
```
smira, line 151: We are using Patch method to avoid conflicts iirc

app/metal-controller-manager/controllers/server_controller.go
Unix4ever: oh, nice, thanks :) This API is a bit in a mess, hard to find things.

app/metal-controller-manager/controllers/server_controller.go
Unix4ever: thanks, fixed.

andrewrynhard:
I would argue that this is a `fix` and should be back ported.

Unix4ever:
> I would argue that this is a fix and should be back ported.

Sure, updated commit msg and title. Regarding back port: should I do something special for that? 
I can probably cherry-pick and create another PR to merge into release.

sfyra/pkg/tests/server_class.go
```diff
@@ -234,6 +235,35 @@ func TestServerClassPatch(ctx context.Context, metalClient client.Client, cluste
...
+		require.NoError(t, metalClient.Delete(ctx, &dummyServer))
+
+		serverStillExists := true
+		labelSelector, err := labels.Parse("dummy-server=")
+		require.NoError(t, err)
+
+		dummyServers := &v1alpha1.ServerList{}
+
+		// try checking that server is still there 10 times
+		retry.Constant(10*time.Second, retry.WithUnits(1*time.Second)).Retry(func() error { // nolint:errcheck
```
smira, line 249: nit: I guess this might be simplified to check server existence in 10 seconds? we don't really need to use retry here, or we could do a simple for loop asserting that object exists on each iteration?

Plus we could use `.Get()` method instead of `List()` to get object by name directly and check that there're no errors.

smira:
> > I would argue that this is a fix and should be back ported.
> 
> Sure, updated the commit msg and the title. Regarding back port: should I do something special for that?
> I can probably cherry-pick and create another PR to merge into the release.

Make sure you're using `git cherry-pick -x` when backporting to `release-0.1` branch so that it shows the original commit id

smira (approved):

PR #268, opened 2021-01-13
docs: add IPMI info, fix links, update environment

smira:
Update environment kernel args and links to the latest Talos release,
add IPMI info, fix random broken links.

### DIFF
--- a/docs/website/content/docs/v0.1/Configuration/servers.md
+++ b/docs/website/content/docs/v0.1/Configuration/servers.md
@@ -85,3 +85,27 @@ You should never change an accepted `Server` to be _not_ accepted while it is in
 Because servers which are not accepted will not be modified, if a server which
 _was_ accepted is changed to _not_ accepted, the disk will _not_ be wiped upon
 its exit.
+
+## IPMI
+
+Sidero can use IPMI information to control `Server` power state, reboot servers and set boot order.
+
+IMPI connection information can be set in the `Server` spec after initial registration:
+
+```yaml
+apiVersion: metal.sidero.dev/v1alpha1
+kind: Server
+...
+spec:
+  bmc:
+    endpoint: 10.0.0.25
+    user: admin
+    pass: password
+```
+
+If IPMI information is set, server boot order might be set to boot from disk, then network, Sidero will switch servers
+to PXE boot once that is required.
+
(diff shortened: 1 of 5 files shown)


docs/website/content/docs/v0.1/Configuration/servers.md
```diff
@@ -85,3 +85,27 @@ You should never change an accepted `Server` to be _not_ accepted while it is in
...
+
+## IPMI
+
+Sidero can use IPMI information to control `Server` power state, reboot servers and set boot order.
+
+IMPI connection information can be set in the `Server` spec after initial registration:
+
+```yaml
+apiVersion: metal.sidero.dev/v1alpha1
+kind: ServerClass
```
nberlee, line 97: I think you mean `kind: Server` here, as `ServerClass` does not have `bmc` according to `kubectl explain serverclass.spec`

andrewrynhard: Yep, this is correct. Good catch.

smira: thanks for catching that!

andrewrynhard (approved):

sfyra/pkg/capi/cluster.go
```diff
@@ -89,6 +89,14 @@ func NewCluster(ctx context.Context, metalClient runtimeclient.Reader, clusterNa
...
 
 		for _, machine := range machines.Items {
+			if !machine.DeletionTimestamp.IsZero() {
+				continue
+			}
+
+			if v1alpha3.MachinePhase(machine.Status.Phase) != v1alpha3.MachinePhaseRunning {
+				continue
+			}
+
```
andrewrynhard, line 99: This the fix for the worker scaling flakes?

smira: yes, yes, and yes!

andrewrynhard: Nice 😄

smira: I got fed up restarting the build, so I went ahead and fixed it hopefully.

Basically machine was marked as deleted, replicas in the status of machinedeployment is accurate, but we picked up stale info by looking into deleted machine

PR #283, opened 2021-02-27
Use LabelSelector for ServerClass filtering.

bzub:
Adds a `LabelSelector` field to the `Qualifiers` struct.

TODO:
- [ ] Depreciate `labelSelectors`?
- [x] Tests
- [ ] Docs

Fixes https://github.com/talos-systems/sidero/issues/282

### DIFF
--- a/app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
+++ b/app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
@@ -4,7 +4,38 @@
 
 package v1alpha1
 
-import "sort"
+import (
+	"sort"
+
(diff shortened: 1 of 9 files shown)


andrewrynhard:
@bzub We had an internal discussion about this. Unfortunately it will break current users, but we think we might be able to cut a `v1alpha2` config version. The migration is simple enough in that case. Gives us a bit to chew on it. We will be starting to look at Sidero again in the coming week or so.

Ulexus:
If we simply _don't_ remove the old LabelSelectors (with an s), mark it as deprecated, and just _add_ the new LabelSelector (no 's'), we shouldn't break anything.

bzub:
Closing in favor of https://github.com/talos-systems/sidero/pull/411

PR #292, opened 2021-03-31
feat: add support for DNS resolution in the agent

andrewrynhard:
This symlinks `/proc/net/pnp` to `/etc/resolv.conf` so that DNS
resolution will work in the agent.


bzub:
https://github.com/talos-systems/sidero/issues/290

smira (approved):

smira:
This seems to fail consistently, I would assume that the change is not okay.

smira:
I think I know what the problem is, tesitng a fix

rsmitty (approved):

PR #300, opened 2021-04-07
fix: don't marshal the machine configuration via machinery package

### LINKED ISSUE #295: Sidero metadata server drops "unknown" fields during config patching
Sidero built against e.g. machinery package from Talos 0.8 drops fields introduced in Talos 0.9

smira:
This fixes the issue with config fields in the machine configuration
introduced in new versions of Talos being lost when configuration is
delivered via the Sidero metadata server.

Instead of marshaling config using machinery, build a config patch to be
applied to fix up kubelet extra args.

P.S. extraArgs structure is really hard to use with configPatches today
as it can be seen from this PR, there's no way to build "easy" patch, as
patch depends on existing extraArgs value.

Fixes #295


rsmitty (approved):
❤️

PR #301, opened 2021-04-08
fix: break the potential endless reconcile loop of ServerClasses

smira:
As list of server names (UUIDs) was not sorted, and listing and
filtering servers is not deterministic, so on each ServerClass reconcile
ServerClass might be changed which triggers another reconcile and so on
until list of server UUIDs on two reconciles stays the same.


AlekSi (approved):
:shipit:

PR #318, opened 2021-04-21
feat: add an option to reference IPMI creds via the secret refs

smira:
v0.3 docs are a plain copy of v0.2 plus changed the
`Configuration/servers.md` for this feature.


smira:
I don't know a good way to test this with Sfyra, but I did manual test verifying that it launches `ipmitool` with correct args

rsmitty (approved):

PR #320, opened 2021-04-22
feat: pull the Sidero configuration as `clusterctl` variables

smira:
This allows to skip patching any components on initial install, and if
the variables are preserved, settings are preserved across upgrades.

### DIFF
--- a/docs/website/content/docs/v0.3/Guides/bootstrapping.md
+++ b/docs/website/content/docs/v0.3/Guides/bootstrapping.md
@@ -123,38 +123,14 @@ As of Cluster API version 0.3.9, Sidero is included as a default infrastructure 
 To install Sidero and the other Talos providers, simply issue:
 
 ```bash
-clusterctl init -b talos -c talos -i sidero
+SIDERO_METADATA_SERVER_HOST_NETWORK=true \
+  SIDERO_METADATA_SERVER_PORT=9091 \
+  SIDERO_CONTROLLER_MANAGER_HOST_NETWORK=true \
+  SIDERO_CONTROLLER_MANAGER_API_ENDPOINT=$PUBLIC_IP \
+  clusterctl init -b talos -c talos -i sidero
 ```
-
-## Patch Components
-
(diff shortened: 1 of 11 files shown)


docs/website/content/docs/v0.3/Guides/bootstrapping.md
```diff
@@ -123,38 +123,14 @@ As of Cluster API version 0.3.9, Sidero is included as a default infrastructure
 To install Sidero and the other Talos providers, simply issue:
 
 ```bash
-clusterctl init -b talos -c talos -i sidero
+SIDERO_METADATA_SERVER_HOST_NETWORK=true \
```
andrewrynhard, line 126: Indent issues?

smira: I think we can indent it this way, can't we? if you prefer other style, I can format it other way

andrewrynhard: I don't care. Just looked like a typo to me at first glance. The subsequent lines are indented and that threw me off.

smira: I was trying to make it clear it's a single command, if all of them have same indentation, it might look like multiple commands

rsmitty (approved):

andrewrynhard (approved):
Environment variables feel like a weak way to handle configuration, but I don't have any alternative ideas. Maybe a configuration file that is specific to Sidero.

smira:
> Environment variables feel like a weak way to handle configuration, but I don't have any alternative ideas. Maybe a configuration file that is specific to Sidero.

`clusterctl` supports setting them in the configuration file as well. Just hard to put that as an example here in the docs, but basically one can define them once and they will be retained around upgrades/installs of the component.

rsmitty:
There's some technical reasons why it has to be this way, unfortunately, if we want to be able to support clusterctl upgrade/move commands.

PR #323, opened 2021-04-23
docs: add guide for upgrades

rsmitty:
This PR points to existing Talos upgrade docs and details the cert
migration for v0.8 -> v0.9 of Talos

### DIFF
--- a/docs/website/content/docs/v0.2/Guides/upgrades.md
+++ b/docs/website/content/docs/v0.2/Guides/upgrades.md
@@ -1,0 +1,68 @@
+---
+description: "A guide describing upgrades"
+weight: 3
+---
+
+# Upgrading
+
+Upgrading a running workload cluster or management plane is the same process as describe in the Talos documentation. 
+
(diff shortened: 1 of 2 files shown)


docs/website/content/docs/v0.2/Guides/upgrades.md
```diff
@@ -0,0 +1,68 @@
...
+  crt: xxx
+  key: xxx
+k8sserviceaccount:
+  key: xxx
+os:
+  crt: xxx
+  key: xxx
+```
+
+- Re-encode the data with `cat <saved-file> | base64 | tr -d '\n'`
```
smira, line 57: nit: `base64 -w 0` so that `tr` is not needed, could be as simple as `base64 -w 0 <saved-file>`

smira (approved):
:+1:

docs/website/content/docs/v0.2/Guides/upgrades.md
rsmitty: Yeah the base64 command that ships with Mac doesn't have that flag, so I always just use trim to make it x-platform

PR #354, opened 2021-04-27
feat: embed iPXE script into the iPXE binaries

### LINKED ISSUE #227: Sidero should construct custom iPXE images
Sidero should construct custom iPXE images for each environment, referencing Sidero's metadata server URL.  This eliminates the loop problem of bootstrapping iPXE and then loading the iPXE config without resorting to class distinctions from the DHCP server, making Sidero much easier to implement on more networks.

See the iPXE chainloading [page](https://ipxe.org/howto/chainloading), Breaking the Infinite Loop with an embedded script section.

Note that the EMBED mechanism works for both undionly.kpxe (BIOS) and ipxe.efi (EFI) iPXE bootstraps (even though the documentation does not say so).

It is further suggested that Sidero may accept user-supplied versions, which may have additional site-specific information.  This kind of override could, perhaps, be supplied in a Secret  (Sidero generates a default one or uses one, if it already exists).

Secrets/ConfigMaps would be the natural place to store these generated binary blobs, since metadata information could be added, and they could easily be overridden.  However, the default ipxe.efi image is perilously close to the 1MB limit, so it seems the use of a Volume (default local) is the only reasonable option.  It is worth noting, however, that this file does compress very well.  I show better than 50% compression with `gzip -9`.

So the idea would be to store in a Volume something like:

```
/environment/cluster-name-0/amd64/ipxe.efi
/environment/cluster-name-0/amd64/undionly.ipxe
/environment/cluster-name-0/amd64/url-hash.undionly.ipxe.md5
/environment/cluster-name-0/amd64/url-hash.ipxe.efi.md5
/environment/cluster-name-0/aarch64/ipxe.efi
/environment/cluster-name-0/aarch64/url-hash.ipxe.efi.md5
```

Th url-hash.md5 files are arbitrarily-named, but are to quickly determine whether the existing binaries were built against the same Sidero metadata URL as what it "currently" the case (thus indicating whether they need to be rebuilt).  Some other scheme is perfectly fine.

## building

In general, the configuration consists of a file named, say, `talos.ipxe`, put in the `src/` directory of the ipxe [source tree](https://ipxe.org/download).  This file would have the format of something like:

```
#!ipxe
dhcp
chain http://sidero.ip.address:8081/boot.ipxe
```

Then, the images are built:
```sh
make bin-x86_64-efi/ipxe.efi EMBED=talos.ipxe
make bin/undionly.kpxe EMBED=talos.ipxe
```

These builds should be able to be spawned by Jobs, so they can have their own short-lived build containers.

smira:
Fixes #227

See also https://github.com/talos-systems/pkgs/pull/267

The rough idea is that iPXE binaries we're building contain placeholder
script which is replaced on the fly with the actual script which depends
on Sidero API endpoint. Patching is a bit tricky for `undionly.kpxe`, as
it is compressed, so we patch uncompressed version and compress the
final asset after that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,10 @@ FROM --platform=arm64 ghcr.io/talos-systems/linux-firmware:${PKGS} AS pkg-linux-
 FROM ghcr.io/talos-systems/musl:${PKGS} AS pkg-musl
 FROM --platform=amd64 ghcr.io/talos-systems/kernel:${PKGS} AS pkg-kernel-amd64
 FROM --platform=arm64 ghcr.io/talos-systems/kernel:${PKGS} AS pkg-kernel-arm64
+FROM ghcr.io/talos-systems/liblzma:${PKGS} AS pkg-liblzma
+FROM ghcr.io/talos-systems/ipxe:${PKGS} AS pkg-ipxe
+FROM --platform=amd64 ghcr.io/talos-systems/ipxe:${PKGS} AS pkg-ipxe-amd64
+FROM --platform=arm64 ghcr.io/talos-systems/ipxe:${PKGS} AS pkg-ipxe-arm64
 
 # The base target provides the base for running various tasks against the source
 # code
@@ -122,11 +126,6 @@ ARG TARGETARCH
 RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=${TARGETARCH} go build -ldflags "-s -w" -o /manager ./app/metal-controller-manager
 RUN chmod +x /manager
 
-FROM scratch AS assets
-ADD http://boot.ipxe.org/undionly.kpxe /amd64/undionly.kpxe
-ADD http://boot.ipxe.org/ipxe.efi /amd64/ipxe.efi
-ADD http://boot.ipxe.org/arm64-efi/ipxe.efi /arm64/ipxe.efi
-
 FROM base AS agent-build-amd64
 RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=amd64 go build -ldflags "-s -w" -o /agent ./app/metal-controller-manager/cmd/agent
 RUN chmod +x /agent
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -149,21 +148,23 @@ COPY --from=pkg-linux-firmware-arm64 /lib/firmware/bnx2 ./lib/firmware/bnx2
 COPY --from=pkg-linux-firmware-arm64 /lib/firmware/bnx2x ./lib/firmware/bnx2x
 RUN set -o pipefail && find . 2>/dev/null | cpio -H newc -o | xz -v -C crc32 -0 -e -T 0 -z >/initramfs.xz
 
-FROM scratch AS metal-controller-manager
+FROM scratch AS metal-controller-manager-image
```
smira, line 151: split this up to squash layers in the final image

Dockerfile
```diff
@@ -149,21 +148,23 @@ COPY --from=pkg-linux-firmware-arm64 /lib/firmware/bnx2 ./lib/firmware/bnx2
...
 COPY --from=pkg-libressl / /
+COPY --from=pkg-liblzma / /
 COPY --from=pkg-ipmitool / /
-COPY --from=assets /amd64/undionly.kpxe /var/lib/sidero/tftp/undionly.kpxe
-COPY --from=assets /amd64/undionly.kpxe /var/lib/sidero/tftp/undionly.kpxe.0
-COPY --from=assets /amd64/ipxe.efi /var/lib/sidero/tftp/ipxe.efi
-COPY --from=assets /arm64/ipxe.efi /var/lib/sidero/tftp/ipxe-arm64.efi
+COPY --from=pkg-ipxe-amd64 /usr/libexec/ /var/lib/sidero/ipxe/amd64
+COPY --from=pkg-ipxe-arm64 /usr/libexec/ /var/lib/sidero/ipxe/arm64
+COPY --from=pkg-ipxe /usr/libexec/zbin/zbin /bin/zbin
```
smira, line 160: copy `zbin` of native architecture to the image

docs/website/content/docs/v0.3/Guides/bootstrapping.md
```diff
@@ -61,14 +61,7 @@ allow bootp;
...
 next-server 192.168.1.150;
-if exists user-class and option user-class = "iPXE" {
-  filename "http://192.168.1.150:8081/boot.ipxe";
-} elsif substring (option vendor-class-identifier, 0, 10) = "HTTPClient" {
-  option vendor-class-identifier "HTTPClient";
-  filename "http://192.168.1.150:8081/tftp/ipxe.efi";
-} else {
-  filename "ipxe.efi";
-}
+filename "ipxe.efi";
```
smira, line 64: this is the gist of the whole PR - `if`s are not longer required

sfyra/pkg/vm/set.go
```diff
@@ -161,7 +161,7 @@ func (set *Set) create(ctx context.Context) error {
 				},
 				PXEBooted:           true,
 				TFTPServer:          set.options.BootSource.String(),
-				IPXEBootFilename:    fmt.Sprintf("http://%s:8081/boot.ipxe", set.options.BootSource),
+				IPXEBootFilename:    "undionly.kpxe",
```
smira, line 164: we test new embedded script with Sfyra - despite the fact QEMU has built-in iPXE, we replace it with our own

Ulexus (approved):

rsmitty (approved):

PR #371, opened 2021-05-05
feat: enable auto-setup of BMC

### LINKED ISSUE #353: Automatically discover and configure BMC (IPMI)
Using `ipmitool` via the Sidero agent, discover if the IPMI (BMC) is available on the node.

If it is available, do the following:

* discover its endpoint (IP address) and send it back as part of the server registration response
* if automatic account creation is enabled, create account `SIDERO` with some random password, send it back as part of the server registration response so that server can fill out credentials (storing password as Kubernetes secret)

rsmitty:
This PR will enable the ability to automatically configure a sidero user
in the BMC if we can complete the process. It removes the need for users
to configure this info per-server if enabled.

Fixes #353 


app/metal-controller-manager/cmd/agent/main.go
```diff
@@ -349,3 +364,150 @@ func mainFunc() error {
...
+
+		return nil
+	})
+
+	return nil
+}
+
+// Returns a random pass string of len 16
+func genPass16() string {
+	rand.Seed(time.Now().UnixNano())
```
smira, line 503: as this function is potentially sensitive to the randomness of the data, and time might be not initialized properly (though unlikely), we might benefit from kernel level random source which should always be better, e.g. with equivalent function [crypto/rand.Int](https://pkg.go.dev/crypto/rand#Int)

smira (commented):
:+1:

app/metal-controller-manager/cmd/agent/main.go
```diff
@@ -349,3 +365,154 @@ func mainFunc() error {
...
+	summResp, err := ipmiClient.GetUserSummary()
+	if err != nil {
+		return err
+	}
+
+	maxUsers := summResp.MaxUsers & 0x1F // Only bits [0:5] provide this number
+
+	// Check if sidero user already exists by combing through all userIDs
+	// nb: we start looking at user id 2, because 1 should always be an unamed admin user and
+	//     we don't want to confuse that unnamed admin with an open slot we can take over.
```
andrewrynhard, line 406: Good job on the comments. Explains the _why_ clearly. 👍🏻

andrewrynhard (approved):
My only question is: do we want to use the new secret credentials option instead of saving directly to the server resource?

rsmitty:
Oh, probably. I hadn't really considered that that was ready for use :man_facepalming:

app/metal-controller-manager/cmd/agent/main.go
rsmitty: I think I've got all of your nits taken care of in amended commit. I want to look at the secret for creds first before merging though.

andrewrynhard:
> Oh, probably. I hadn't really considered that that was ready for use 🤦‍♂️

The good news is that I think it will be relatively easy. But that is easy for me to say I suppose 😄 .

smira:
> > Oh, probably. I hadn't really considered that that was ready for use man_facepalming
> 
> The good news is that I think it will be relatively easy. But that is easy for me to say I suppose  .

we might need to figure out a strategy for that -- create a single secret for all servers?

app/metal-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -34,13 +35,19 @@ type BMC struct {
...
+	// +optional
+	Interface string `json:"interface,omitempty"`
 }
 
 // CredentialSource defines a reference to the credential value.
 type CredentialSource struct {
-	// Selects a key of a secret in the cluster namespace
-	// +optional
-	SecretKeyRef *corev1.SecretKeySelector `json:"secretKeyRef,omitempty"`
+	// Namespace and name of credential secret
```
rsmitty, line 45: Keep SecretKeyRef and squirrel this stuff under that.

app/metal-controller-manager/internal/server/server.go
```diff
@@ -258,7 +273,89 @@ func (s *server) Heartbeat(ctx context.Context, in *api.HeartbeatRequest) (*api.
...
+			if !apierrors.IsNotFound(err) {
+				return nil, err
+			}
+
+			log.Printf("BMC secret doesn't exist for server %q, creating.", in.GetUuid())
+
+			credsSecret.ObjectMeta = metav1.ObjectMeta{
+				Namespace: "sidero-system",
+				Name:      in.GetUuid() + "-bmc",
+				OwnerReferences: []metav1.OwnerReference{
```
rsmitty, line 300: Does clusterctl move move this resource?

app/metal-controller-manager/config/manager/manager.yaml
```diff
@@ -64,6 +64,7 @@ spec:
             - --extra-agent-kernel-args=${SIDERO_CONTROLLER_MANAGER_EXTRA_AGENT_KERNEL_ARGS:=-}
             - --auto-accept-servers=${SIDERO_CONTROLLER_MANAGER_AUTO_ACCEPT_SERVERS:=false}
             - --insecure-wipe=${SIDERO_CONTROLLER_MANAGER_INSECURE_WIPE:=true}
+            - --auto-bmc-setup=${SIDERO_CONTROLLER_MANAGER_AUTO_BMC_SETUP:=true}
```
rsmitty, line 67: add to docs

app/metal-controller-manager/internal/server/server.go
rsmitty: They should indeed get moved, but there's some nuance to it since secrets are namespaced. I'm going to move back to creating these secrets in the default namespace since that's where our clusters are also created by default and thus the secrets would get picked up properly with a `clusterctl move`

app/metal-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -49,13 +61,16 @@ func (source *CredentialSource) Resolve(ctx context.Context, reader client.Clien
 		return "", nil
 	}
 
-	if source.SecretKeyRef == nil {
```
smira, line 52: I believe this check should stay, as one might not have the secret key ref (by mistake?)

app/metal-controller-manager/internal/server/server.go
```diff
@@ -94,16 +95,30 @@ func (s *server) CreateServer(ctx context.Context, in *api.CreateServerRequest)
 		log.Printf("Added %s", in.GetSystemInformation().GetUuid())
 	}
 
-	resp := &api.CreateServerResponse{}
+	resp := &api.CreateServerResponse{
+		SetupBmc: false,
```
smira, line 99: nit: we can remove that, as it's default (zero) value

app/metal-controller-manager/internal/server/server.go
```diff
@@ -94,16 +95,30 @@ func (s *server) CreateServer(ctx context.Context, in *api.CreateServerRequest)
...
+
+	// Make BMC and wiping decisions only if server is accepted
+	// to avoid hijacking random devices that PXE boot against us.
+	if obj.Spec.Accepted {
+		// Respond to agent whether it should attempt bmc setup
+		// We will only tell it to attempt autoconfig if there's not already data there.
+		if obj.Spec.BMC == nil && s.autoBMC {
+			log.Printf("Server %q needs BMC setup", obj.Name)
+
+			resp.SetupBmc = s.autoBMC
```
smira, line 110: nit: we check `s.autoBMC` above, so we might as well set it to `true` here

app/metal-controller-manager/internal/server/server.go
```diff
@@ -258,7 +273,94 @@ func (s *server) Heartbeat(ctx context.Context, in *api.HeartbeatRequest) (*api.
...
+		// Fetch corresponding server
+		obj := &metalv1alpha1.Server{}
+
+		if err := s.c.Get(ctx, types.NamespacedName{Name: in.GetUuid()}, obj); err != nil {
+			return nil, err
+		}
+
+		// For auto-created BMC info, we will *always* drop the creds in default namespace
+		// This ensures they'll come along for the ride in a "default" clusterctl move based on our cluster templates.
+		if err := s.c.Get(ctx, types.NamespacedName{Namespace: corev1.NamespaceDefault, Name: in.GetUuid() + "-bmc"}, credsSecret); err != nil {
```
smira, line 291: nit: `in.GetUuid() + "-bmc"` might benefit from being extracted as a variable, as we use it many times in this function

docs/website/content/docs/v0.3/Configuration/servers.md
```diff
@@ -77,7 +77,7 @@ Sidero will never write to or wipe any disk on a computer which is not marked as
 This can be tedious for systems in which all attached computers should be considered to be under the control of Sidero.
 Thus, you may also choose to automatically accept any machine into Sidero on its discovery.
 Please keep in mind that this means that any newly-connected computer **WILL BE WIPED** automatically.
-You can enable auto-acceptance by pasing the `--auto-accept-servers=true` flag to `sidero-controller-manager`.
+You can enable auto-acceptance by passing the `--auto-accept-servers=true` flag to `sidero-controller-manager`.
```
smira, line 80: we should also document the new variable in the `Geting Started/installation.md` document

rsmitty: alrighty @smira, think I've got your feedback taken care of.

smira (approved):
:v:

PR #373, opened 2021-05-06
feat: ship ServerClass "any"

### LINKED ISSUE #310: Ship a default ServerClass which matches all Servers
The simplest case is that Sidero is given Just a Bunch of Servers, and they are all basically equivalent to the user.  This is common for small test setups and for larger buildouts which occur all at once.  Regardless of the reason, for maximum simplicity and utility, we should have a default ServerClass which matches _all_ Servers, and we should ship that with Sidero as the default.

AlekSi:
Closes #310.


app/metal-controller-manager/api/v1alpha1/server_types_test.go
```diff
@@ -58,19 +58,6 @@ func Test_PartialEqual(t *testing.T) {
...
-				a: v1alpha1.CPUInformation{
-					Manufacturer: "QEMU",
-				},
-				b: v1alpha1.CPUInformation{
-					Manufacturer: "QEMU",
-					Version:      "1.2.0",
-				},
-			},
-			want: true,
-		},
```
AlekSi, line 73: That a full copy of another test above

app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
```diff
@@ -0,0 +1,112 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package v1alpha1
+
+import "sort"
+
+// FilterAcceptedServers returns a new slice of Servers that are accepted and qualify.
+//
+// Returned Servers are always sorted by name for stable results.
+func FilterAcceptedServers(servers []Server, q Qualifiers) []Server {
```
AlekSi, line 12: Converted the previous version of this code to use pure functions instead of _both_ mutating an internal state and returning it

app/metal-controller-manager/controllers/serverclass_controller.go
```diff
@@ -30,127 +30,32 @@ type ServerClassReconciler struct {
...
 // +kubebuilder:rbac:groups=metal.sidero.dev,resources=servers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=metal.sidero.dev,resources=servers/status,verbs=get;update;patch
 
 func (r *ServerClassReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
+
 	l := r.Log.WithValues("serverclass", req.NamespacedName)
+	l.Info("reconciling")
+
+	//nolint:godox
```
AlekSi, line 44: I think it should be fine to add TODOs to our code without this directive

app/metal-controller-manager/controllers/serverclass_controller.go
```diff
@@ -30,127 +30,32 @@ type ServerClassReconciler struct {
...
+	// TODO: We probably should use admission webhooks instead (or in additional) to prevent
+	// unwanted edits instead of "fixing" the resource after the fact.
+	if req.Name == metalv1alpha1.ServerClassAny {
+		if err := ReconcileServerClassAny(ctx, r.Client); err != nil {
+			return ctrl.Result{}, err
+		}
 
-	l.Info("fetching serverclass", "serverclass", req.NamespacedName)
+		// do not return; re-reconcile it to update status
+	} //nolint:wsl
```
AlekSi, line 53: The code above looks like a typical Go code with `return` at the end of the block. The comment explains why `return` is not there, that it is not forgotten. `wsl` linter's warning is… unfortunate.

app/metal-controller-manager/main.go
```diff
@@ -193,6 +195,17 @@ func main() {
...
+	k8sClient, err := client.NewClient(nil)
+	if err != nil {
+		setupLog.Error(err, `failed to create k8s client`)
+		os.Exit(1)
+	}
+
+	if err = controllers.ReconcileServerClassAny(context.TODO(), k8sClient); err != nil {
+		setupLog.Error(err, `failed to reconcile ServerClass "any"`)
+		os.Exit(1)
+	}
```
AlekSi, line 207: We have to use a separate client there due to controller-runtime's caching "subsystem" is not running yet.

… Actually, I should use msg.GetAPIReader() I just found instead :(

hack/start-registry-proxies.sh
```diff
@@ -28,8 +28,3 @@ docker run -d -p 5004:5000 \
     -e REGISTRY_PROXY_REMOTEURL=https://ghcr.io \
     --restart always \
     --name registry-ghcr.io registry:2
-
-docker run -d -p 5004:5000 \
-    -e REGISTRY_PROXY_REMOTEURL=https://ghcr.io \
-    --restart always \
-    --name registry-ghcr.io registry:2
```
AlekSi, line 35: A duplicate of the previous block

rsmitty (approved):

app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, server := range servers {
+		if server.Spec.Accepted {
+			res = append(res, server)
+		}
+	}
+
+	return res
+}
+
+func filterCPU(servers []Server, filters []CPUInformation) []Server {
```
smira, line 35: not for this PR, but I think we should refactor this to do one pass through the slice applying all the conditions to avoid copying and iterating over the list many times

app/metal-controller-manager/main.go
smira: one more idea to handle this is to inject reconcilation for `any` at the beginning of the controller work

smira: I don't have any _awesome_ idea about it, but something in `SetupWithManager` should do the trick imho

AlekSi: I actually tried to do that in `SetupWithManager` – cache is still not running at that point

app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
AlekSi: I will do that in this PR as I already refactored this code

smira (dismissed):

smira (approved):

PR #375, opened 2021-05-07
chore: allow building with debug handlers

AlekSi:
Refs talos-systems/talos#3534.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,7 +71,7 @@ RUN protoc -I/src/app/metal-controller-manager/internal/api \
   --go_out=paths=source_relative:/src/app/metal-controller-manager/internal/api --go-grpc_out=paths=source_relative:/src/app/metal-controller-manager/internal/api \
   api.proto
 RUN --mount=type=cache,target=/.cache controller-gen object:headerFile="./hack/boilerplate.go.txt" paths="./..."
-RUN --mount=type=cache,target=/.cache	conversion-gen --input-dirs="./app/cluster-api-provider-sidero/api/v1alpha2" --output-base ./ --output-file-base="zz_generated.conversion" --go-header-file="./hack/boilerplate.go.txt"
+RUN --mount=type=cache,target=/.cache conversion-gen --input-dirs="./app/cluster-api-provider-sidero/api/v1alpha2" --output-base ./ --output-file-base="zz_generated.conversion" --go-header-file="./hack/boilerplate.go.txt"
 ARG MODULE
 RUN --mount=type=cache,target=/.cache gofumports -w -local ${MODULE} .
 
@@ -108,7 +108,8 @@ COPY --from=release-build /cluster-template.yaml /infrastructure-sidero/${TAG}/c
 
 FROM base AS build-cluster-api-provider-sidero
 ARG TARGETARCH
-RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=${TARGETARCH} go build -ldflags "-s -w" -o /manager ./app/cluster-api-provider-sidero
+ARG GO_BUILDFLAGS
+RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=${TARGETARCH} go build ${GO_BUILDFLAGS} -ldflags "-s -w" -o /manager ./app/cluster-api-provider-sidero
 RUN chmod +x /manager
 
 ## TODO(rsmitty): make bmc pkg and move to talos-systems image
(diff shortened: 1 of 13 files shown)


smira:
this all sounds pretty reasonable to me, we probably need to keep debug.enabled as a generic build tag? 

so :+1:

andrewrynhard:
All for making it generic, though profiling should be disabled by default for security reasons.

AlekSi:
"sidero" in "sidero.debug" is a company name, not a product name :) To keep our tags namespaced

Dockerfile
```diff
@@ -71,7 +71,7 @@ RUN protoc -I/src/app/metal-controller-manager/internal/api \
   --go_out=paths=source_relative:/src/app/metal-controller-manager/internal/api --go-grpc_out=paths=source_relative:/src/app/metal-controller-manager/internal/api \
   api.proto
 RUN --mount=type=cache,target=/.cache controller-gen object:headerFile="./hack/boilerplate.go.txt" paths="./..."
-RUN --mount=type=cache,target=/.cache	conversion-gen --input-dirs="./app/cluster-api-provider-sidero/api/v1alpha2" --output-base ./ --output-file-base="zz_generated.conversion" --go-header-file="./hack/boilerplate.go.txt"
+RUN --mount=type=cache,target=/.cache conversion-gen --input-dirs="./app/cluster-api-provider-sidero/api/v1alpha2" --output-base ./ --output-file-base="zz_generated.conversion" --go-header-file="./hack/boilerplate.go.txt"
```
AlekSi, line 74: There was a tab character instead of space

Makefile
```diff
@@ -16,6 +16,14 @@ PKGS ?= v0.5.0-8-gb0d9cd2
 
 SFYRA_CLUSTERCTL_CONFIG ?= $(HOME)/.cluster-api/clusterctl.sfyra.yaml
 
+GO_BUILDFLAGS ?=
+
+WITH_DEBUG ?=
+
+ifeq ($(WITH_DEBUG), true)
```
AlekSi, line 23: It will not work with `WITH_DEBUG=0` or `WITH_DEBUG=1`. I'm not fond of it at all. It is very tempting to write a boolean parser in `hack/parsebool.(py|sh)` and invoke it there via shell.

app/cluster-api-provider-sidero/main.go
```diff
@@ -26,6 +28,10 @@ import (
 	// +kubebuilder:scaffold:imports
 )
 
+const (
+	debugAddr = ":9994"
```
AlekSi, line 32: Port numbers are sequential but basically in the random range.

Thanks to `debug.Enabled` being a constant, port number can be configurable via flags, environment variables, etc. Should it be?

app/metal-controller-manager/config/crd/bases/metal.sidero.dev_servers.yaml
```diff
@@ -71,7 +71,7 @@ spec:
                     description: Source for the password value. Cannot be used if Pass is not empty.
                     properties:
                       secretKeyRef:
-                        description: SecretKeyRef defines a ref to a given key within a secret
+                        description: SecretKeyRef defines a ref to a given key within a secret.
```
AlekSi, line 74: It was missing at #371, and our CI did not catch it

app/metal-metadata-server/main.go
```diff
@@ -60,8 +74,9 @@ func main() {
 		client: k8sClient,
 	}
 
-	http.HandleFunc("/configdata", mm.FetchConfig)
```
AlekSi, line 63: Default server mux shouldn't be used there because debug handlers will use it

internal/client/client.go
```diff
@@ -19,7 +19,7 @@ import (
 	metal "github.com/talos-systems/sidero/app/metal-controller-manager/api/v1alpha1"
 )
 
-// NewClient is responsible for creating a controller-runtime k8s client for use by the metadata server.
+// NewClient is responsible for creating a controller-runtime k8s client with all schemas.
```
AlekSi, line 22: Missed from the other PR

Makefile
smira: another option is to check whether `WITH_DEBUG` is not empty, but that might be overkill

app/cluster-api-provider-sidero/main.go
smira: I think hardcoded should be fine?

app/metal-controller-manager/config/crd/bases/metal.sidero.dev_servers.yaml
smira: I thought this gets generated from the `server_types.go`, so I think _believe_ it should be changed there

smira (dismissed):

Makefile
AlekSi: Check for empty string will enable debug build with `WITH_DEBUG=0`, and that's even worse

app/metal-controller-manager/config/crd/bases/metal.sidero.dev_servers.yaml
AlekSi: That's due to missing `check_dirty` pipeline step

Unix4ever (dismissed):

PR #392, opened 2021-05-13
release(v0.3.0-alpha.0): prepare release

rsmitty:
This is the official v0.3.0-alpha.0 release.

### DIFF
--- a/hack/chglog/config.yml
+++ b/hack/chglog/config.yml
@@ -2,15 +2,15 @@ style: github
 template: CHANGELOG.tpl.md
 info:
   title: CHANGELOG
-  repository_url: https://github.com/talos-systems/talos
+  repository_url: https://github.com/talos-systems/sidero
 options:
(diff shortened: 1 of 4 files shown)


hack/chglog/config.yml
```diff
@@ -2,15 +2,15 @@ style: github
 template: CHANGELOG.tpl.md
 info:
   title: CHANGELOG
-  repository_url: https://github.com/talos-systems/talos
+  repository_url: https://github.com/talos-systems/sidero
```
smira, line 5: not for this PR, but I would love us to switch to `release-tool` like Talos

smira (dismissed):

hack/chglog/config.yml
rsmitty: oh yeah, totally. I had some trouble with the release script this time. will add that as a 0.3 task.

PR #403, opened 2021-05-17
fix: check all labels in a serverclass labelselector

bzub:
We can provide zero or more labelSelectors in a ServerClass to
filter Servers. Each labelSelector is a map of labels. This fixes
a bug where if only one label in one labelSelector matches, all
servers with that label would match.

https://github.com/talos-systems/sidero/issues/288


bzub:
Decided not to touch `labelSelectors` and implement https://github.com/talos-systems/sidero/issues/282 instead.

PR #408, opened 2021-05-18
feat: ship Environment "default"

### LINKED ISSUE #311: Ship a default Environment
Most of an Envirtonment can be automatically determined, and even for customizations, it is easier to modify an existing template than to start from scratch.

We should supply a default boot Environment with Sidero; mostly, this just involves knowing Sidero's own endpoint name and the "current" Talos release.

AlekSi:
Closes #311.

Signed-off-by: Alexey Palazhchenko <alexey.palazhchenko@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -121,9 +121,10 @@ LABEL org.opencontainers.image.source https://github.com/talos-systems/sidero
 ENTRYPOINT [ "/manager" ]
 
 FROM base AS build-metal-controller-manager
+ARG TALOS_RELEASE
 ARG TARGETARCH
 ARG GO_BUILDFLAGS
-RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=${TARGETARCH} go build ${GO_BUILDFLAGS} -ldflags "-s -w" -o /manager ./app/metal-controller-manager
+RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=${TARGETARCH} go build ${GO_BUILDFLAGS} -ldflags "-s -w -X main.TalosRelease=${TALOS_RELEASE}" -o /manager ./app/metal-controller-manager
 RUN chmod +x /manager
 
 FROM base AS agent-build-amd64
--- a/app/metal-controller-manager/api/v1alpha1/environment_types.go
+++ b/app/metal-controller-manager/api/v1alpha1/environment_types.go
@@ -5,8 +5,15 @@
 package v1alpha1
 
 import (
+	"fmt"
+	"sort"
+
+	"github.com/talos-systems/talos/pkg/machinery/kernel"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// EnvironmentDefault is an automatically created Environment.
+const EnvironmentDefault = "default"
 
 type Asset struct {
 	URL    string `json:"url,omitempty"`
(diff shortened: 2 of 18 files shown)


Dockerfile
```diff
@@ -121,9 +121,10 @@ LABEL org.opencontainers.image.source https://github.com/talos-systems/sidero
 ENTRYPOINT [ "/manager" ]
 
 FROM base AS build-metal-controller-manager
+ARG TALOS_RELEASE
 ARG TARGETARCH
 ARG GO_BUILDFLAGS
-RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=${TARGETARCH} go build ${GO_BUILDFLAGS} -ldflags "-s -w" -o /manager ./app/metal-controller-manager
+RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=${TARGETARCH} go build ${GO_BUILDFLAGS} -ldflags "-s -w -X main.TalosRelease=${TALOS_RELEASE}" -o /manager ./app/metal-controller-manager
```
smira, line 127: :heart:

app/metal-controller-manager/api/v1alpha1/environment_types.go
```diff
@@ -65,6 +71,42 @@ type EnvironmentList struct {
 	Items           []Environment `json:"items"`
 }
 
+// EnvironmentDefaultSpec returns EnvironmentDefault's spec.
+func EnvironmentDefaultSpec(talosRelease, endpoint string) *EnvironmentSpec {
+	return &EnvironmentSpec{
+		Kernel: Kernel{
+			Asset: Asset{
+				URL: fmt.Sprintf("https://github.com/talos-systems/talos/releases/download/%s/vmlinuz-%s", talosRelease, runtime.GOARCH),
```
smira, line 79: I would rather put `arch` as `amd64` for now, as our support for provisioning `arm64` boxes is pretty weak right now, while it's perfectly fine to provision `amd64` boxes from Sidero running on Raspberry Pi.

app/metal-controller-manager/api/v1alpha1/environment_types.go
```diff
@@ -65,6 +71,42 @@ type EnvironmentList struct {
...
+			},
+			Args: []string{
+				"console=tty0",
+				"console=ttyS1,115200n8",
+				"consoleblank=0",
+				"earlyprintk=ttyS1,115200n8",
+				"ima_appraise=fix",
+				"ima_hash=sha512",
```
smira, line 87: Major part of these args can be sourced from https://github.com/talos-systems/talos/blob/master/pkg/machinery/kernel/kernel.go

app/metal-controller-manager/api/v1alpha1/environment_types.go
```diff
@@ -65,6 +71,42 @@ type EnvironmentList struct {
...
+				"init_on_free=1",
+				"initrd=initramfs.xz",
+				"page_poison=1",
+				"printk.devkmsg=on",
+				"pti=on",
+				"random.trust_cpu=on",
+				"slab_nomerge",
+				"slub_debug=P",
+				"talos.platform=metal",
+				fmt.Sprintf("talos.config=http://%s:8081/configdata?uuid=", endpoint),
```
smira, line 99: we have now a variable for `apiPort` as well which we should use here instead of 8081

smira (approved):
:partying_face:

PR #411, opened 2021-05-18
feat: add label selector to serverclass

### LINKED ISSUE #282: Use LabelSelector instead of map
We should use the standard Kubernetes [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector) for the ServerClass label-matching scheme.:

```yaml
apiVersion: metal.sidero.dev/v1alpha1
kind: ServerClass
metadata:
  name: example
spec:
  qualifiers:
    cpu:
      - manufacturer: Intel(R) Corporation
        version: Intel(R) Atom(TM) CPU C3558 @ 2.20GHz
    # old deprecated method under qualifiers; one OR other element should match; somewhat broken
    labelSelectors:
      - {foo: bar, baz: qux}
      - {quux: quuz}
  # new method under spec
  selector:
    matchLabel:
      corge: grault
    matchExpressions: # the requirements are ANDed
      - {key: garply, operator: In, values: [waldo, fred]}
      - {key: plugh, operator: In, values: [xyzzy]}
```

If both `qualifiers` and a new `selector` are present, both should match. That is, `qualifiers AND selector`. If one of them is empty, it is assumed to be `true` (`true AND selector`, `qualifiers AND true`). If both are empty, this ServerClass matches all Servers (`true AND true`); that's what ServerClass `any` does.

bzub:
Adds field `selector` to the root of the ServerClassSpec.

This implements the same schema/logic for selecting Servers as is
used in several builtin Kubernetes APIs.

Tests have been added to show the behavior when both `selector` and
`qualifiers` are used, as well as when neither are used.

https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/label-selector/

Signed-off-by: bzub <bzubrod@gmail.com>

Closes #282

### DIFF
--- a/app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
+++ b/app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
@@ -4,19 +4,37 @@
 
 package v1alpha1
 
-import "sort"
+import (
+	"fmt"
+	"sort"
 
-// FilterAcceptedServers returns a new slice of Servers that are accepted and qualify.
-//
-// Returned Servers are always sorted by name for stable results.
-func FilterAcceptedServers(servers []Server, q Qualifiers) []Server {
-	res := make([]Server, 0, len(servers))
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
 
-	for _, server := range servers {
-		// skip non-accepted servers
-		if !server.Spec.Accepted {
-			continue
+// AcceptedServerFilter matches Servers that have Spec.Accepted set to true.
+func AcceptedServerFilter(s Server) (bool, error) {
+	return s.Spec.Accepted, nil
+}
+
+// SelectorFilter returns a ServerFilter that matches servers against the
+// serverclass's selector field.
+func (sc *ServerClass) SelectorFilter() func(Server) (bool, error) {
+	return func(server Server) (bool, error) {
+		s, err := metav1.LabelSelectorAsSelector(&sc.Spec.Selector)
+		if err != nil {
+			return false, fmt.Errorf("failed to get selector from labelselector: %v", err)
 		}
+
+		return s.Matches(labels.Set(server.GetLabels())), nil
+	}
+}
+
+// QualifiersFilter returns a ServerFilter that matches servers against the
+// serverclass's qualifiers field.
+func (sc *ServerClass) QualifiersFilter() func(Server) (bool, error) {
+	return func(server Server) (bool, error) {
+		q := sc.Spec.Qualifiers
 
 		// check CPU qualifiers if they are present
 		if filters := q.CPU; len(filters) > 0 {
@@ -30,7 +48,7 @@ func FilterAcceptedServers(servers []Server, q Qualifiers) []Server {
 			}
 
 			if !match {
-				continue
+				return false, nil
 			}
 		}
 
@@ -45,7 +63,7 @@ func FilterAcceptedServers(servers []Server, q Qualifiers) []Server {
 			}
 
 			if !match {
-				continue
+				return false, nil
 			}
 		}
 
@@ -62,14 +80,41 @@ func FilterAcceptedServers(servers []Server, q Qualifiers) []Server {
 			}
 
 			if !match {
-				continue
+				return false, nil
 			}
 		}
 
-		res = append(res, server)
+		return true, nil
+	}
+}
+
+// FilterServers returns the subset of servers that pass all provided filters.
+// In case of error the returned slice will be nil.
+func FilterServers(servers []Server, filters ...func(Server) (bool, error)) ([]Server, error) {
+	matches := make([]Server, 0, len(servers))
+
+	for _, server := range servers {
+		match := true
+
+		for _, filter := range filters {
+			var err error
+
+			match, err = filter(server)
+			if err != nil {
+				return nil, fmt.Errorf("failed to filter server: %v", err)
+			}
+
+			if !match {
+				break
+			}
+		}
+
+		if match {
+			matches = append(matches, server)
+		}
 	}
 
-	sort.Slice(res, func(i, j int) bool { return res[i].Name < res[j].Name })
+	sort.Slice(matches, func(i, j int) bool { return matches[i].Name < matches[j].Name })
 
-	return res
+	return matches, nil
 }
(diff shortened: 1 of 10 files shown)


bzub:
`make lint` succeeded on my laptop before I rebased with today's commits. Hm.

andrewrynhard:
> `make lint` succeeded on my laptop before I rebased with today's commits. Hm.

I am fixing this in my PR. Sorry about that. It's a mistake on our end.

andrewrynhard:
@bzub Linting is fixed in `master` now.

website/content/docs/v0.1/Configuration/serverclasses.md
```diff
@@ -5,29 +5,60 @@ weight: 3
 
 # Server Classes
 
-Server classes are a way to group distinct server resources.
-The "qualifiers" key allows the administrator to specify criteria upon which to group these servers.
-There are currently three keys: `cpu`, `systemInformation`, and `labelSelectors`.
-Each of these keys accepts a list of entries.
-The top level keys are a "logical AND", while the lists under each key are a "logical OR".
-Qualifiers that are not specified are not evaluated.
+Server classes are a way to group distinct server resources.  The `qualifiers`
```
andrewrynhard, line 8: We run linters that should be enforcing https://sembr.org. Not sure why this isn't caught, but fair game if it is broken. Just FYI. TL;DR is that there should be a sentence per-line. I am not going to block on it since our linting isn't catching it.

andrewrynhard (commented):
Other than integration tests failing, this LGTM.

app/metal-controller-manager/api/v1alpha1/serverclass_types.go
```diff
@@ -22,6 +22,7 @@ type Qualifiers struct {
 type ServerClassSpec struct {
 	EnvironmentRef *corev1.ObjectReference `json:"environmentRef,omitempty"`
 	Qualifiers     Qualifiers              `json:"qualifiers"`
+	Selector       *metav1.LabelSelector   `json:"selector"`
```
AlekSi, line 25: It seems that `metav1.LabelSelector` is typically used as a value field, not as a pointer field. `metav1.LabelSelectorAsSelector` will convert empty non-nil LabelSelector to `labels.Everything()` which is what we need, right? That should simplify some logic.

app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
```diff
@@ -4,7 +4,38 @@
...
+		if !server.Spec.Accepted {
+			continue
+		}
+
+		if sc.Spec.Selector == nil || s.Matches(labels.Set(server.GetLabels())) {
+			matches = append(matches, server)
+		}
+	}
+
+	return FilterAcceptedServers(matches, sc.Spec.Qualifiers), nil
```
AlekSi, line 37: We should un-export this method if it is no longer used directly.

website/content/docs/v0.1/Configuration/serverclasses.md
```diff
@@ -5,29 +5,60 @@ weight: 3
...
+## `selector`
+
+`selector` groups server resources by their labels. The [Kubernetes
+documentation][label-selector-docs] has more information on how to use this
+field.
+
+## `qualifiers`
+
+There are currently four keys: `cpu`, `systemInformation`, `labelSelectors`,
+and `labelSelector`.  Each of these keys accepts a list of entries.  The top
```
smira, line 23: `labelSelector` key is not there, and I think we should stop documenting `labelSelectors` to make it less confusing. 

whoever is using `labelSelectors`, they will keep using it, but we don't want new users of this field.

website/content/docs/v0.1/Configuration/serverclasses.md
```diff
@@ -5,29 +5,60 @@ weight: 3
...
+          - prod
   qualifiers:
     cpu:
-      - manufacturer: Intel(R) Corporation
-        version: Intel(R) Atom(TM) CPU C3558 @ 2.20GHz
-      - manufacturer: Advanced Micro Devices, Inc.
-        version: AMD Ryzen 7 2700X Eight-Core Processor
-    labelSelectors:
-      - "my-server-label": "true"
+      - manufacturer: "Intel(R) Corporation"
```
smira, line 50: should we leave two CPUs here as an example (as it was before with AMD CPU)?

website/content/docs/v0.1/Configuration/serverclasses.md
```diff
@@ -5,29 +5,60 @@ weight: 3
...
+        version: "Intel(R) Atom(TM) CPU C3558 @ 2.20GHz"
+    systemInformation:
```
smira, line 52: this doesn't look like a good example to me with all fields being `Unknown` :)

app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
smira: and we should probably rename these methods as it's a bit confusing now 

`Accepted` is now filtered twice in two functions, we could have high-level exported `FilterServers` which calls a chain of match functions for each server. 

```go

for _, server := range servers {
  matches := true
  
  for _, filter := range []FilterFunction{isAccepted,selectorMatches,qualifiersMatch, ... } {
     if !filter(server) {
         matches = false
         break
     }
     
     if matches {
         result = append(result, server)
      }
}
```

website/content/docs/v0.1/Configuration/serverclasses.md
AlekSi: https://github.com/talos-systems/sidero/pull/422

AlekSi: That PR is merged now, linting should catch that

app/metal-controller-manager/api/v1alpha1/serverclass_types.go
bzub: Good point. Fixed.

website/content/docs/v0.1/Configuration/serverclasses.md
bzub: Agreed. Fixed.

app/metal-controller-manager/api/v1alpha1/serverclass_filter.go
bzub: Implemented something like this. Thanks.

website/content/docs/v0.1/Configuration/serverclasses.md
bzub: Fixed the style. Thanks.

smira (approved):
:+1:

PR #413, opened 2021-05-18
feat: provide several options to force boot from disk via iPXE

### LINKED ISSUE #287: Make ipxe behavior customizable
One knob I need is to customize the ipxe response when a server is allocated/installed. Sidero currently returns an ipxe script that executes `exit` in the hopes of booting from disk, but our servers halt booting if this happens, requiring manual intervention. I would change this to returning an HTTP error code if I had such an option, which our servers then try other boot options or reboot as needed.

In short, I'd like to be able to change this:
```
❯ curl -s 'http://assets.management:8081/ipxe?uuid=UUID-OF-INSTALLED-SERVER'
#!ipxe
exit
```

to this:
```
❯ curl -v 'http://assets.management:8081/ipxe?uuid=UUID-OF-INSTALLED-SERVER'
[...]
< HTTP/1.1 404 Not Found
[...]
```

Or something else? `exit` seems to cause issues.

Related: https://ipxe.org/appnote/work_around_bios_halting_on_ipxe_exit

smira:
If the server is configured to PXE boot by default, it might hit the
Sidero iPXE server after Talos install, so Sidero has to force the
server to boot from disk.

Sidero supported default method via iPXE `exit` command, but this
command doesn't always work
([details](https://ipxe.org/appnote/work_around_bios_halting_on_ipxe_exit)).

This PR provides two other ways to force to boot from disk:

* `http-404` force HTTP 404 response from iPXE server
* `ipxe-sanboot` uses `sanboot` command to boot from the first disk

This should improve compatibility with some BIOSes.

This setting should eventually be provided per-server as well.

Fixes #287


app/metal-controller-manager/api/v1alpha1/environment_types.go
```diff
@@ -76,7 +76,7 @@ type EnvironmentList struct {
 func EnvironmentDefaultSpec(talosRelease, apiEndpoint string, apiPort uint16) *EnvironmentSpec {
 	args := make([]string, 0, len(kernel.DefaultArgs)+6)
 	args = append(args, kernel.DefaultArgs...)
-	args = append(args, "console=tty0", "console=ttyS1,115200n8", "earlyprintk=ttyS1,115200n8")
```
smira, line 79: this fixes default environment console args. we can't guess whether it's serial port zero or one, but providing zero makes more sense, as it matches Talos defaults on platform `metal` and also fixes Sfyra console output

Unix4ever (approved):

andrewrynhard (approved):

PR #445, opened 2021-05-25
feat: always return BMC IP if found

rsmitty:
This PR makes sure that we're attempting to patch the BMC struct twice.
Once for any IP we find for the BMC info, and a second time to add the
sidero user. This ensures that we'll at least populate the IP if we can.

### DIFF
--- a/app/metal-controller-manager/cmd/agent/main.go
+++ b/app/metal-controller-manager/cmd/agent/main.go
@@ -262,11 +262,20 @@ func mainFunc() error {
 	if createResp.GetSetupBmc() {
 		log.Println("Attempting to automatically discover and configure BMC")
 
-		// nb: we don't consider failure to get BMC info a hard failure.
+		// Attempt to discover the BMC IP
+		// nb: we don't consider failure to get BMC info a hard failure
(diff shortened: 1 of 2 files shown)


rsmitty:
unfortunately, there seems to be no good way to "turn off" the local user abilities to test this out fully. but I can at least say that this successfully creates and updates the BMC info, so no breaking of current functionality.

app/metal-controller-manager/cmd/agent/main.go
```diff
@@ -262,11 +262,20 @@ func mainFunc() error {
...
+		err := attemptBMCIP(ctx, client, s)
+		if err != nil {
+			log.Printf("encountered error discovering BMC IP. skipping setup: %q", err.Error())
+		}
+
+		// Attempt to add sidero user to BMC
 		// nb: we don't consider failure to get BMC info a hard failure.
 		//     users can always patch the bmc info to the server themselves.
-		err := attemptBMCSetup(ctx, client, s)
+		err = attemptBMCUserSetup(ctx, client, s)
```
smira, line 276: should we skip this if BMC IP discovery fails?

smira (dismissed):

PR #495, opened 2021-07-05
chore: update Talos to 0.11.0-beta.2

smira:
Looks like `go` doesn't like pseudo-revisions for replaced same-worktree
folders anymore. I think we should start pushing proper tags for
`pkg/machinery/` and ignore them in the CI.


smira:
waits for #494

AlekSi:
That `replace` shouldn't be needed, will take a look

smira:
> That `replace` shouldn't be needed, will take a look

Feels like something new in go behavior, it complains about fake module version.

I think proper fix is to actually start tagging machinery

AlekSi:
Yeah, we should make a tag `pkg/machinery/v0.11.0-beta.2`

AlekSi (approved):

AlekSi (approved):

PR #499, opened 2021-07-06
fix: make sure powercycle condition gets properly updated

smira:
We rely on the timestamp of the `PowerCycle` condition to be updated
properly when Sidero reboots the node (assuming previous attempt to PXE
boot failed). Looks like nothing we did before actually ensures that the
condition gets properly updated, so a little hack modifying condition
message to make sure actual condition gets patched properly with last
transition time.


smira:
it might make sense to backport this to 0.3

andrewrynhard (approved):

PR #537, opened 2021-08-04
fix: backport fixes

rsmitty:
This PR backports the following fixes:

- fix: update Sfyra to install CAPI v0.3 [commit here](ea3016f)
- fix: update sidero IPMI user to work properly on idrac [commit here](7bdee0f)

### DIFF
--- a/app/sidero-controller-manager/cmd/agent/main.go
+++ b/app/sidero-controller-manager/cmd/agent/main.go
@@ -530,10 +530,10 @@ func attemptBMCUserSetup(ctx context.Context, client api.AgentClient, s *smbios.
 	}
 
 	// Make sidero an admin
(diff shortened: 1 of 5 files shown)


andrewrynhard (approved):

andrewrynhard (approved):
Typically we use `git cherry-pick -x` for each commit we need to back port and just ignore the "maximum of one commit" check. Not a blocker for me.

rsmitty:
yeah, that's probably a better way to do it. I'll redo, hang tight. I actually did that, then squashed but your way is better for sure.

rsmitty:
@andrewrynhard k, fixed. I'll let that build and merge it once passing.

andrewrynhard (approved):

PR #554, opened 2021-08-16
fix: resource config links

frezbo:
Fix redirects to resource configuration


smira (approved):
thank you!

PR #568, opened 2021-09-17
fix: shutdown sidero-controller-manager when any component fails

### LINKED ISSUE #560: Add health checks to all controllers
We should add readiness and liveness problem to all controllers.

smira:
Fixes #560

The way it was implemented before this change, `errgoup` waits for all
goroutines to finish before it returns, so if the controller crashes due
to election issues, container still keeps running as HTTP API is up.

After this change, container crashes on first error.

Also added liveness/readiness check, they won't help much this issue,
but provide additional layer of protection/visibility.


smira:
should be backported to 0.3

Unix4ever (approved):

PR #607, opened 2021-10-20
feat: allow defining http server and api server ports separately

Unix4ever:
This should provide flexibility in case if running on `hostNetwork` and
want a different port and still keep compatibility with
`loadbalancer`/`ingress` setups.


smira:
nit, but might be easier to submit as two PRs, as the docs don't need to be backported

Unix4ever:
cc @rgl we decided to split these port configs into two part so you'll need to additionally specify `SIDERO_CONTROLLER_MANAGER_CONTAINER_API_PORT=80` when we get a new release.

PR #639, opened 2021-11-16
feat: conformance check

frezbo:
Add Conformance check


smira:
@frezbo this should be a `chore:` better, as it doesn't have any user-visible feature

smira (approved):

Issue #623, opened 2021-11-06
Failing to boot RPI4


charlie-haley:
I've set up [Sidero on RPI4 as per the documentation.](https://www.sidero.dev/docs/v0.4/guides/sidero-on-rpi4/), which is working perfectly and my x86 machines have no issue PXE booting into Talos.

[I then followed the guide to configure RPI4 as servers](https://www.sidero.dev/docs/v0.4/guides/rpi4-as-servers/), which works up until the UEFI firmware part. The Pi's boot over TFTP and successfully load the UEFI firmware, but they get stuck attempting to PXE boot.

My default environment file looks like this, I even tried setting up a ServerClass and still no luck
```
---
apiVersion: metal.sidero.dev/v1alpha1
kind: Environment
metadata:
  name: default
  namespace: default
spec:
  kernel:
    url: "https://github.com/talos-systems/talos/releases/download/v0.13.2/vmlinuz-arm64"
    args:
      - console=tty0
      - console=ttyS1,115200n8
      - consoleblank=0
      - earlyprintk=ttyS1,115200n8
      - ima_appraise=fix
      - ima_hash=sha512
      - ima_template=ima-ng
      - init_on_alloc=1
      - initrd=initramfs.xz
      - nvme_core.io_timeout=4294967295
      - printk.devkmsg=on
      - pti=on
      - random.trust_cpu=on
      - slab_nomerge=
      - talos.config=http://192.168.1.215:8081/configdata?uuid=
      - talos.platform=metal
  initrd:
    url: "https://github.com/talos-systems/talos/releases/download/v0.13.2/initramfs-arm64.xz"

```
This is the log from the UEFI firmware - it successfully downloads the file, but then does nothing and just moves onto the next item in the boot order.
```
>>Start PXE over IPv4.
     Station IP address is 192.168.1.8

     Server IP address is 192.168.1.215
     NBP filename is ipxe.efi
     NBP filesize is 1019200 Bytes
    Downloading NBP file...

     NBP file downloaded successfully.
```

Any ideas on what could be causing this?

frezbo:
@charlie-haley I believe the pi is downloading `ipxe.efi` instead of `ipxe-arm64.efi` and failing to execute it since it's the wrong architecture. When I set this up for Pi I had 

```bash
filename "ipxe-arm64.efi"; 
```

set explicitly in the DHCP config. Can you update the dhcp config and try again

charlie-haley:
> 
> 
[...]

PR #640, opened 2021-11-16
docs: dhcp option-93

frezbo:
- Example of multi-arch network boot

Ref: #623


smira (approved):
:+1:

PR #641, opened 2021-11-16
feat: automatically append talos.config to the Environment

### LINKED ISSUE #460: Environment no longer needs talos.config
We create an environment with the `talos.config` arg. This is inject automatically now. We should remove it.

smira:
Fixes #460

With this change, it's no longer necessary to supply `talos.config` in
the Environment - it is appended dynamically during the boot process.

So `talos.config` is removed even from the default Environment, as this
makes it update automatically if Sidero endpoint gets changed.


smira:
Draft as I need to update the docs, but we first need to have 0.5 docs forked.

smira:
image

this was my accidental push to `origin`, PR build is fine

rsmitty (approved):

PR #647, opened 2021-11-24
fix: update CAPI resources versions to v1alpha4 in the cluster template

Unix4ever:
It worked before as it was being migrated from `v1alpha3` to `v1alpha4` when applied.

Fixes: https://github.com/talos-systems/sidero/issues/628


smira (approved):

kurious-diru:
lgtm

PR #648, opened 2021-11-25
feat: support pausing server

frezbo:
Adds `paused` field to *server* spec for maintenance


app/sidero-controller-manager/config/crd/bases/metal.sidero.dev_servers.yaml
```diff
@@ -25,6 +25,10 @@ spec:
       jsonPath: .spec.accepted
       name: Accepted
       type: boolean
+    - description: indicates if the server is paused
```
smira, line 28: I don't know if `paused` is the right word, as it still allows server  to wiped, deallocated, etc. 

It just stops the server from being allocated.

@Ulexus ?

sfyra/pkg/tests/server.go
```diff
@@ -259,6 +259,127 @@ func TestServerAcceptance(ctx context.Context, metalClient client.Client, vmSet
...
+			require.NoError(t, err)
+		}
+
+		dummyServers := &v1alpha1.ServerList{}
+
+		labelSelector, err := labels.Parse("dummy-server=")
+		require.NoError(t, err)
+
+		// wait for all the servers to be registered
+		require.NoError(t, retry.Constant(1*time.Minute, retry.WithUnits(10*time.Second)).Retry(func() error {
```
smira, line 286: I don't think there's a way dummyServers can be not visible immediately after creation?

sfyra/pkg/tests/server.go
```diff
@@ -259,6 +259,127 @@ func TestServerAcceptance(ctx context.Context, metalClient client.Client, vmSet
...
+		}
+
+		// verify that all servers shows up as available in `any` serverclass
+		require.NoError(t, retry.Constant(1*time.Minute, retry.WithUnits(10*time.Second)).Retry(func() error {
+			var serverClass v1alpha1.ServerClass
+			err := metalClient.Get(ctx, types.NamespacedName{Name: v1alpha1.ServerClassAny}, &serverClass)
+			if err != nil {
+				return err
+			}
+			if len(serverClass.Status.ServersAvailable) != numDummies {
```
smira, line 325: this is a wrong check: there might be other `Server`s left from previous runs, so we should better check that `ServerClass` explicitly lists dummies' IDs in the list as 'Available'

website/content/docs/v0.4/Getting Started/create-workload.md
```diff
@@ -35,7 +35,7 @@ For instance:
 ```bash
 export CONTROL_PLANE_SERVERCLASS=any
 export WORKER_SERVERCLASS=any
-export TALOS_VERSION=v0.13.0
+export TALOS_VERSION=v0.13.3
```
smira, line 38: this is _config_ version, not explicitly Talos version iirc, so it doesn't have to be bumped

sfyra/pkg/tests/server.go
frezbo: fixed

sfyra/pkg/tests/server.go
frezbo: fixed

website/content/docs/v0.4/Getting Started/create-workload.md
frezbo: updated so that `make talos-artifacts` downloads the stable. This also fixed the issue with talosctl not able to destroy created machines

frezbo:
accidentally deleted the branch, re-created https://github.com/talos-systems/sidero/pull/649

PR #650, opened 2021-11-26
feat: implement Talos kernel log receiver

### LINKED ISSUE #527: log receiver service
Sidero should offer a logging service which accepts logs from managed machines and stores them to either (configurable) RAM-based ring buffer or disk storage.

These logs should be grouped by Cluster and by Machine.

The design of Loki seems like it fits well with the needs here, and Loki offers some useful storage backends, while being substantially lighter weight than other log aggregators like ElasticSearch.  So perhaps the options should be ring buffer, disk file, or Loki.

In any case, we need both client and server sides of this feature:  Sidero should act as a log server to broker and direct logs.  It should configure Environments and Talos installations to set `netconsole=` kernel commandline flags to its log service.  Thus, no client-side service is required (though we may wish to add in other logs later), just the kernel config.

smira:
Fixes #527

Talos logs (see https://github.com/talos-systems/talos/pull/4600) are
delivered to Sidero over the SideroLink tunnel.

Logs can be seen with:

```
$ kubectl logs -n sidero-system deployment/sidero-controller-manager -c serverlogs -f
{"clock":67194673,"cluster":"management-cluster","facility":"user","machine":"default/management-cluster-cp-4j8f4","metal_machine":"default/management-cluster-cp-hbq57","msg":"[talos] phase bootloader (19/19): done, 176.795226ms\n","priority":"warning","seq":768,"server_uuid":"5b72932a-c482-4aa5-b00e-4b8773d3ac48","talos-level":"warn","talos-time":"2021-11-26T19:34:42.444342392Z"}
```

Logs are annotated on the fly with the information about `Server`,
`MetalMachine`, `Machine` and `Cluster`.


app/sidero-controller-manager/internal/ipxe/ipxe_server.go
```diff
@@ -405,14 +407,21 @@ func newEnvironmentFromServerClass(serverBinding *infrav1.ServerBinding) (env *m
 }
 
 func appendTalosArguments(env *metalv1alpha1.Environment) {
+	if sideroLinkConfig.ServerAddress.IsZero() {
```
Unix4ever, line 410: Maybe can load it in `newEnvironment` call. Then it will be possible to handle an error there.

smira: yeah, it's weird. we can't load before mgr is started, as client is not initialized yet.

we don't have a good pre-hook for it. probably we can put it to some `errgroup` in `main`

smira: I ended using `mgr.Add` in `main.go`

PR #662, opened 2021-12-03
feat: populate addresses and hostname in ServerBindings

Unix4ever:
Get addresses from the events coming into events sink container.

Add v1beta1 support.
Update `metalmachine_controller` to set providerID earlier and add
ProviderSet condition to be set after the cluster nodes actually get
patched with the provider ID as well.

The condition is required to trigger reconcile on CAPI `machines`,
otherwise it never picks `Node` ready condition.


app/caps-controller-manager/api/v1alpha3/metalmachine_conditions.go
```diff
@@ -0,0 +1,20 @@
...
+// +kubebuilder:object:generate=true
+// +groupName=controlplane.cluster.x-k8s.io
+package v1alpha3
+
+import clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+
+// Conditions and condition Reasons for the TalosControlPlane object
+
+const (
+	// ProviderSetCondition reports when the nodes have provider ID label defined.
```
smira, line 14:
```suggestion
	// ProviderSetCondition reports when the nodes have .spec.Provider ID field set.
```

?

app/sidero-controller-manager/cmd/events-manager/adapter.go
```diff
@@ -91,3 +140,77 @@ func (a *Adapter) HandleEvent(ctx context.Context, event events.Event) error {
...
+		}
+	}
+
+	a.nodesMu.Lock()
+	defer a.nodesMu.Unlock()
+
+	if new == nil {
+		delete(a.nodes, oldServerBinding.Spec.SideroLink.NodeAddress)
+	} else {
+		address := newServerBinding.Spec.SideroLink.NodeAddress
```
smira, line 202: if serverBinding got update, we should remove old address the map, to delete unused addresses

hack/release.toml
```diff
@@ -38,4 +44,11 @@ Events:
 ```
 kubectl logs -f -n sidero-system deployment/sidero-controller-manager serverevents
 ```
+"""
+
+    [notes.addresses]
+        title = "Machine Addresses"
+        description = """\
+Sidero now populates `MetalMachine` addresses with the ones discovered from Siderolink server events.
+Which is then propagated to CAPI `Machine` resources.
```
smira, line 53: might need to add here that this feature is only available with Talos >= 0.14

sfyra/cmd/sfyra/cmd/options.go
```diff
@@ -67,7 +67,7 @@ func DefaultOptions() Options {
 		TalosKernelURL: fmt.Sprintf("https://github.com/talos-systems/talos/releases/download/%s/vmlinuz-amd64", TalosRelease),
 		TalosInitrdURL: fmt.Sprintf("https://github.com/talos-systems/talos/releases/download/%s/initramfs-amd64.xz", TalosRelease),
 
-		CoreProvider:            "cluster-api:v0.4.3",
+		CoreProvider:            "cluster-api:v1.0.0",
```
smira, line 70: should we skip locking the version of CAPI here? there's no bug now in CAPI which installed wrong version not matching the contract

templates/cluster-template.yaml
```diff
@@ -52,9 +52,6 @@ spec:
     apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
     name: ${CLUSTER_NAME}-cp
   controlPlaneConfig:
-    init:
```
smira, line 55: :partying_face:

sfyra/go.mod
```diff
@@ -43,8 +43,8 @@ require (
 	k8s.io/apiextensions-apiserver v0.22.3
 	k8s.io/apimachinery v0.23.0-alpha.4
 	k8s.io/client-go v0.23.0-alpha.4
-	sigs.k8s.io/cluster-api v0.4.3
```
smira, line 46: we can probably now try dropping the "hold" block above which holds some versions back to preserve compatibility

sfyra/cmd/sfyra/cmd/options.go
Unix4ever: If all 1.x.x will be backward compatible then yeah.

smira (approved):

PR #667, opened 2021-12-08
feat: extend information printed in the iPXE script, add retries

smira:
With this change, iPXE now prints information about available network
interfaces and configured IP after DHCP.

It should also retry DHCP but it's hard to test in my environment.

Also different interface should be tried if loading `/ipxe` script
fails.

Example:

```
net0: 4e:65:61:9e:84:a4 using undionly on 0000:00:02.0 (closed)
  [Link:up, TX:0 TXE:1 RX:0 RXE:0]
  [TXE: 1 x "Network unreachable (http://ipxe.org/28086011)"]
Configuring (net0 4e:65:61:9e:84:a4)...... ok
net0: 172.25.0.3/255.255.255.0 gw 172.25.0.1
http://172.24.0.2:8081/ipxe... ok
/env/agent-amd64/vmlinuz... ok
```

### DIFF
--- a/app/sidero-controller-manager/internal/ipxe/ipxe_server.go
+++ b/app/sidero-controller-manager/internal/ipxe/ipxe_server.go
@@ -40,11 +40,57 @@ const bootFile = `#!ipxe
 chain ipxe?uuid=${uuid}&mac=${mac:hexhyp}&domain=${domain}&hostname=${hostname}&serial=${serial}&arch=${buildarch}
 `
 
-// bootTemplate is embedded into iPXE binary when that binary is sent to the node.
-//
-// bootTemplate should be kept in sync with the bootFile above.
-var bootTemplate = template.Must(template.New("iPXE embedded").Parse(`dhcp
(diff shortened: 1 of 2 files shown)


app/sidero-controller-manager/internal/ipxe/ipxe_server.go
```diff
@@ -43,8 +43,15 @@ chain ipxe?uuid=${uuid}&mac=${mac:hexhyp}&domain=${domain}&hostname=${hostname}&
 // bootTemplate is embedded into iPXE binary when that binary is sent to the node.
 //
 // bootTemplate should be kept in sync with the bootFile above.
-var bootTemplate = template.Must(template.New("iPXE embedded").Parse(`dhcp
-chain http://{{ .Endpoint }}:{{ .Port }}/ipxe?uuid=${uuid}&mac=${mac:hexhyp}&domain=${domain}&hostname=${hostname}&serial=${serial}&arch=${buildarch}
+var bootTemplate = template.Must(template.New("iPXE embedded").Parse(`
+ifstat
+
+:retry
+dhcp || goto retry
```
smira, line 50: might need to add `ifclose` here before `dhcp`, but it's not clear?

rsmitty (approved):

PR #680, opened 2021-12-14
fix: ignore BMC info if username or password are not set

smira:
This happens during initial discovery of the server, when first BMC IP
is set, but username and password are not set.


app/sidero-controller-manager/internal/power/factory.go
```diff
@@ -39,6 +39,11 @@ func NewManagementClient(ctx context.Context, client client.Client, spec *v1alph
 			}
 		}
 
+		if bmcSpec.User == "" || bmcSpec.Pass == "" {
+			// no username and password, BMC information is not fully populated yet
```
andrewrynhard, line 43: I have very little context, but do we/should we log this?

andrewrynhard (approved):

app/sidero-controller-manager/internal/power/factory.go
smira: I don't think so, it's same as incomplete BMC info, so Sidero won't do power operations with the machine. As we don't log the error when BMC info is not available.

Today with Sidero, it auto-provisions both IP and username/password, but it does that in two transactions, so there's a state when BMC IP is set, but username/password is not. In this case Sidero tries to access IPMI with empty username and password, that fails and leaves bad taste.

rsmitty (approved):

frezbo (approved):

PR #705, opened 2022-01-12
feat: set server PXEBooted condition only after Talos gets installed

Unix4ever:
Now `PXEBooted` condition is derived from `TalosInstalled` condition of
the linked `ServerBinding`.
This makes the node to use `pxe` boot until Talos installation succeeds.

### DIFF
--- a/app/sidero-controller-manager/internal/ipxe/ipxe_server.go
+++ b/app/sidero-controller-manager/internal/ipxe/ipxe_server.go
@@ -220,7 +220,15 @@ func ipxeHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// This code is left here only for backward compatibility with Talos <= v0.13.
 	if !strings.HasPrefix(env.ObjectMeta.Name, "agent") {
+		// Do not mark as PXE booted here if SideroLink events are available and Talos installation is in progress.
+		// SideroLink events handler will mark the machine with TalosInstalledCondition condition,
+		// then server controller will reconcile this status and mark server as PXEBooted.
+		if conditions.Has(serverBinding, infrav1.TalosInstalledCondition) {
+			return
+		}
+
 		if err = markAsPXEBooted(server); err != nil {
 			log.Printf("error marking server as PXE booted: %s", err)
 		}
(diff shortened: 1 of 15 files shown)


app/sidero-controller-manager/internal/ipxe/ipxe_server.go
```diff
@@ -219,12 +217,6 @@ func ipxeHandler(w http.ResponseWriter, r *http.Request) {
 
 		return
 	}
-
-	if !strings.HasPrefix(env.ObjectMeta.Name, "agent") {
```
smira, line 223: I feel this is tough, but we need to find a solution here:

* if Talos <= 0.13, then we won't have any conditions on ServerBinding/MetalMachine, so we should keep doing our "dump" PXE booted if reached out to the PXE server 
* if Talos >= 0.14, the conditions will be present and we can use those vs. what we have right now

I don't know how to implement that fully, but at least with Talos <= 0.13 I think this code will never mark it as PXE booted?

smira: On other hand, PXE booting all the time if the server is configured to boot from network first is not the end of the world.

It's fine and we can live with that. It will break upgrades if someone does an upgrade, and might make big confusion if Environment gets updated, while installed Talos version is different.

smira: What I was thinking about is the following:

* if Talos >= 0.14, and Talos boots, it should send events and we should have _some_ conditions set on the ServerBinding from events; in that case we should stop PXE booting when install succeeds
* if Talos <= 0.13, we would have no conditions set on ServerBinding, but we have this "weak" condition that the server reached out to the iPXE, and we can keep using that?

Unix4ever: yeah... I forgot about backward compatibility here...
I wonder if we can detect if Talos <= 0.13 is used in Sidero somehow.

smira: The only idea I have is the above ^^: judge by the presence of conditions, by the time it gets to boot once again, it should have _some_ conditions set.

Unix4ever: Ok, I think then I need to update event handler to set conditions earlier

smira (approved):

PR #735, opened 2022-02-01
feat: add extended hardware information to Server and ServerClass CRDs

lion7:


app/sidero-controller-manager/api/v1alpha1/server_types.go
```diff
@@ -125,36 +175,57 @@ func PartialEqual(a, b interface{}) bool {
...
-	SystemInformation *SystemInformation      `json:"system,omitempty"`
-	CPU               *CPUInformation         `json:"cpu,omitempty"`
-	BMC               *BMC                    `json:"bmc,omitempty"`
-	ManagementAPI     *ManagementAPI          `json:"managementApi,omitempty"`
-	ConfigPatches     []ConfigPatches         `json:"configPatches,omitempty"`
-	Accepted          bool                    `json:"accepted"`
-	Cordoned          bool                    `json:"cordoned,omitempty"`
-	PXEBootAlways     bool                    `json:"pxeBootAlways,omitempty"`
+	EnvironmentRef *corev1.ObjectReference `json:"environmentRef,omitempty"`
+	Hardware       *HardwareInformation    `json:"hardware,omitempty"`
```
smira, line 221: this is a breaking change for the CRDs, as the fields were removed. We could do also migration and new CRD version, but that might need more changes. I wonder if we could still keep things same way in the CRDs, even if we restructure the fields in Go structs.

lion7: Yeah, that is also what I was struggling with. I agree not breaking things is better, but I honestly don't know a good way.
A partial migration should be possible I think, although it will probably always lead to incomplete information (for example servers with multiple CPU's will only have 1 listed). Also we would only have a few fields of the new set available.
I'm not sure what is worse: not having information or having incomplete information...

frezbo: :+1: for a new CRD version and we do CRD migrations, it'll be less visible changes for a user

smira: I agree we should probably do new API version and migration, but this means we should also fix the webhooks for this controller (it's right now broken/missing) and go all the way with proper migration. And if we do that, should we do more changes as well?

smira:
@lion7 we really like the proposed changes here, but as we would need to do CRD migration, we would like to delay that for the next release of Sidero (0.6.0). We need to fix the webhooks first in the controllers and implement proper migration

lion7:
Yeah, I totally agree. Is there a timeline on the next release? Like in the order of days / weeks / months?

Anyway, I'll see if I can improve this PR by implementing the proposed changes.

lion7:
I moved my CRD changes to a new version and implement the manual conversions were needed. This should not break things I hope. 

I am not entirely sure about the whole conversion part though. The conversions for the `caps-controller-manager` also have these `func (src *MetalCluster) ConvertTo(dstRaw conversion.Hub) error {...}` and `func (dst *MetalCluster) ConvertFrom(srcRaw conversion.Hub) error {...}` functions which I didn't yet implement for the `sidero-controller-manager`. 
Nothing seems to break when building, is this something that's used only at runtime inside a K8s cluster? (haven't tested that yet)

lion7:
I also added network interfaces to the collected information, so we have the complete set of data from servers.

@smira there are some (un)related changes in this PR that I'd like to separate in other PR's so the diff will be smaller.
For example, I had to update all imports due to the new v1alpha2 CRD's and found a lot of places with inconsistent naming for imports. I have moved most of these changes into https://github.com/siderolabs/sidero/pull/793.

I also want to move the Kustomize config changes to a different PR, there are quite a few of them in order to get the whole webhook stuff working. Still working on that one.

lion7:
Finally got the conversion webhooks working 🥳 
I duplicated these changes in https://github.com/siderolabs/sidero/pull/794, so they can be merged independently.
After https://github.com/siderolabs/sidero/pull/793, https://github.com/siderolabs/sidero/pull/794 and https://github.com/siderolabs/sidero/pull/795 are merged I'll update this PR so the diff will only cover the actual changes of adding the extended hardware information as the title suggests.

smira:
@lion7 I want to say once again thank you for your awesome job on this one. Even in the test environment the reporting of hardware information is much improved!

Sorry that it took too long to merge this, but we got these changes merged with some fixups in #823 and #824

I haven't made any changes to the actual CRDs, but rather some small fixups here and there, so I believe this PR might be closed unless I missed something on the way :)

lion7:
@smira great to hear that this feature was merged! No problem on the delay, it was a huge change (and PR) after all 😅.
It all looks good to me, I'll do a local build and test it out on my setup. Closing this PR 🥳.

PR #739, opened 2022-02-03
refactor: cleanup and deduplicate the code which matches SideroLink IPs

smira:
Use same matcher in both log receiever and event receiver. Fix panic in
event receiver.

Bump CAPI and remove cert-manager temporary hack.


smira:
this also properly annotates event logs

andrewrynhard (approved):

PR #749, opened 2022-02-07
chore: upgrade cluster-api version from v0.4.4 to v0.4.7 in getting started guide

tvories:
Let me know if there are any other files I should be updating.  This resolves an installer error with v0.4.4


smira:
thanks for putting that, I will probably make the wording more clear that one should use the latest clusterctl release, as Cluster API is doing releases on their own schedule

tvories:
That totally makes sense. I was copying and pasting the code blocks as my installer.  I knew to use the latest 0.4.x, but the copy and paste obviously wasn't the latest.  I figured at least updating the code block wasn't a bad idea

smira:
Thanks, I incorporated the changes into #752 !

PR #794, opened 2022-03-26
feat: fix conversion webhooks

lion7:
The configs for Sidero and CAPS are processed in a single kustomize call, which caused some resources to conflict. These conflicts are resolved and conversion webhooks should work now.


lion7:
I moved the whole webhook server to a separate binary / image to prevent an chicken and egg problem. The reconcilers of the `sidero-controller-manager` tried to call the webhook service to convert resources, but previously the webhook server was part of the same binary.

Since the `caps-controller-manager` also needs the same webhook service for conversions, I figured it would be cleaner to just move it all into a separate, dedicated container. I tested these changes in combination with https://github.com/siderolabs/sidero/pull/735 and I saw the `Server` CRDs being properly converted.

Unix4ever:
Thanks for your PR. I've fixed that as part of other PR based on yours. Will close this one

PR #795, opened 2022-03-26
feat: allow configuring Sidero deployment strategy

lion7:
Upgrading Sidero on a single-node cluster with host networking enabled fails because the used ports can only be bound once. By setting the deployment strategy to `Recreate` in these cases upgrades can succeed (but with a small amount of downtime).


frezbo (approved):

lion7:
The sfyra build fails with `Error: failed to get provider components for the "talos" provider: target namespace can't be defaulted. Please specify a target namespace`, which is caused by the GitHub rename.
Maybe add a clusterctl.yaml to the CI build? Or just wait for upstream to merge https://github.com/kubernetes-sigs/cluster-api/pull/6328

smira (commented):
:heart: thank you!

PR #812, opened 2022-04-07
feat: enable webhooks in `sidero-controller-manager`

Unix4ever:
Based on https://github.com/siderolabs/sidero/pull/794, but instead of
using a dedicated controller for all webhooks keep all of them in each
own controller.

Additionally, implement validation webhooks for `Server` resource, which
validates that `bootFromDiskMethod` and `configPatches` are using proper
values.

### DIFF
--- a/app/sidero-controller-manager/main.go
+++ b/app/sidero-controller-manager/main.go
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	infrav1alpha3 "github.com/talos-systems/sidero/app/caps-controller-manager/api/v1alpha3"
+	"github.com/talos-systems/sidero/app/sidero-controller-manager/api/v1alpha1"
 	metalv1alpha1 "github.com/talos-systems/sidero/app/sidero-controller-manager/api/v1alpha1"
 	"github.com/talos-systems/sidero/app/sidero-controller-manager/controllers"
 	"github.com/talos-systems/sidero/app/sidero-controller-manager/internal/ipxe"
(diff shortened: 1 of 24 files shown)


smira (approved):
do webhooks work for the `caps-controller-manager` as well?

smira:
> > do webhooks work for the caps-controller-manager as well?
> 
> They should, but I haven't really verified that.

It might be nice to verify that, e.g. with some validation once again, even if that's temporary.

Unix4ever:
> > > do webhooks work for the caps-controller-manager as well?
> > 
> > They should, but I haven't really verified that.
>
> It might be nice to verify that, e.g. with some validation once again, even if that's temporary.

Yep, there actually was an issue with them. I've fixed that.

app/sidero-controller-manager/main.go
```diff
@@ -216,6 +217,7 @@ func main() {
 		os.Exit(1)
 	}
 
+	setupWebhooks(mgr)
```
lion7, line 220: I still don't get how this would work. In https://github.com/siderolabs/sidero/pull/794 I used a dedicated controller for the webhooks because the reconcilers above this line are actively using the conversion webhooks. 

To my understanding this will cause Sidero to be unable to start when any of the CRD's change like proposed in https://github.com/siderolabs/sidero/pull/735, because in that case the reconcilers will try to convert them using the conversion webhooks defined here (but at that point, they are not yet available).

smira: @lion7 I'm going to take a closer look today at your PR and do more testing for the conversion

lion7: 👍🏻 that would be great, in that case I'll wait with rebasing https://github.com/siderolabs/sidero/pull/735 so it won't accidentally break (I have a working deployment of it running on my cluster).

PR #814, opened 2022-04-08
docs: add info about GitHub Org rename to v0.5 install docs

JJGadgets:
While attempting to follow the v0.5 install instructions currently put up on the Sidero website, I encountered the target namespace error, and had to search further to find the solution in https://github.com/siderolabs/cluster-api-bootstrap-provider-talos/blob/master/README.md#installing

This PR adds the info to the Sidero v0.5 install docs, which to my knowledge the files in the repo is directly rendered by the https://sidero.dev website.


JJGadgets:
Seems like there's a formatting process that I didn't follow resulting in the automated tests failing, how would I go about fixing this?

rsmitty:
hey @JJGadgets thanks for adding this. don't sweat it, I'll need to check this out and sign the commit anyways, so I can fix it.

PR #821, opened 2022-04-12
refactor: remove pre-ServerBinding leftovers, use remote cached clients

smira:
This PR has two major changes:

* retire the code which supported seamless migration from
pre-ServerBinding era to ServerBindings: creating `ServerBinding` on the
fly from the `MetalMachine` and `Server`; as there's no migration path
from pre-ServerBinding Sidero to the new version, it's time to drop it
* instead of creating workload cluster Kubernetes client each time, use
CAPI standard class to cache the client; the problem with "leaking"
clients is that HTTP/2 clients are almost never gc'ed, so they stay in
memory keeping an open connection with keepalives going both ways, so
caching lowers the load both on the controller and the control plane
endpoint

### DIFF
--- a/app/caps-controller-manager/controllers/metalmachine_controller.go
+++ b/app/caps-controller-manager/controllers/metalmachine_controller.go
@@ -12,16 +12,14 @@ import (
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/tools/reference"
 	"k8s.io/utils/pointer"
 	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/controllers/remote"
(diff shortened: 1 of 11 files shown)


app/caps-controller-manager/controllers/metalmachine_controller.go
```diff
@@ -133,7 +132,14 @@ func (r *MetalMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	controllerutil.AddFinalizer(metalMachine, infrav1.MachineFinalizer)
 
-	// If server ref is already provided, server binding controller is going to reconcile matching server binding
+	// TODO (smira):
+	// This is really weird that .Spec.ServerRef is used both to set the manual link to the server by the user
+	// and a way to store current binding when server is picked up automatically
+	//
+	// This should be refactored to keep _current_ binding ref in the .Status, while
+	// .spec should be reserved for manually chosen server only.
+	// This opens a question what to do if the `.Status` is lost after pivoting, how to reconcile it correctly?
```
frezbo, line 141: is it possible to tell capi to migrate that over?

frezbo, line 141: is it possible to tell capi to migrate that over?

frezbo (approved):

app/caps-controller-manager/controllers/metalmachine_controller.go
smira: that would be possible if we change the schema, for this PR I just documented the problem as it is right now

frezbo: makes me wonder if we should discuss how the whole servers and serverclasses and the binding gets populated, feels like room for improvement

smira: yeah, I think in Sidero 0.6 we should do more improvements as we change our schema actually

PR #823, opened 2022-04-14
feat: restructure HW information in `Server` resources (v1alpha2)

smira:
This is basically subset of PR #735 with only CRD changes without actual
code changes to use new CRDs. As storage version is v1alpha2, every
access goes twice via conversion webhooks (for better test coverage).

Other parts of #735 will be incorporated in a follow-up PR, I decided to
split things up for easier review.

Example:

```bash
$ kubectl get servers.v1alpha1.metal.sidero.dev 49fd7c2d-1ba4-4157-8cc0-3f7212f119f0 -o yaml
...
  cpu:
    manufacturer: QEMU
    version: pc-q35-6.0
  hostname: pxe-3
  managementApi:
    endpoint: 172.25.0.1:39565
  system:
    family: Unknown
    manufacturer: QEMU
    productName: Standard PC (Q35 + ICH9, 2009)
    serialNumber: Unknown
    skuNumber: Unknown
    version: pc-q35-6.0
$ kubectl get servers.v1alpha2.metal.sidero.dev 49fd7c2d-1ba4-4157-8cc0-3f7212f119f0 -o yaml
...
  hardware:
    compute:
      processors:
      - manufacturer: QEMU
        productName: pc-q35-6.0
    system:
      family: Unknown
      manufacturer: QEMU
      productName: Standard PC (Q35 + ICH9, 2009)
      serialNumber: Unknown
      skuNumber: Unknown
      version: pc-q35-6.0
```

We can make more changes to v1alpha2 resources in follow-up PRs.

Co-authored-by: Gerard de Leeuw <gdeleeuw@leeuwit.nl>
Signed-off-by: Gerard de Leeuw <gdeleeuw@leeuwit.nl>
Signed-off-by: Andrey Smirnov <andrey.smirnov@talos-systems.com>

PR #824, opened 2022-04-14
feat: add extended hardware information to Server and ServerClass CRDs

smira:
This is final part of #735, previous part in #823.

This imports all remaining changes with some fixups minus the webhook
changes.

This change adds detailed hardware information to the Server CRD.
Hardware info is extracted by the agent from SMBIOS.
The ServerClass CRD is also updated so more precise qualifiers can be used.

Co-authored-by: Gerard de Leeuw <gdeleeuw@leeuwit.nl>


app/sidero-controller-manager/config/samples/metal_v1alpha2_environment.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: metal.sidero.dev/v1alpha2
+kind: Environment
+metadata:
+  name: default
+spec:
+  kernel:
+    url: "https://github.com/talos-systems/talos/releases/download/v0.10.3/vmlinuz-amd64"
```
frezbo, line 7: probably just use latest

templates/cluster-template-talos-v0.13.4.yaml
```diff
@@ -38,7 +38,7 @@ spec:
   template:
     spec:
       serverClassRef:
-        apiVersion: metal.sidero.dev/v1alpha1
+        apiVersion: metal.sidero.dev/v1alpha2
```
frezbo, line 41: should be bump to v1alpha2 for the talos v0.13.4 template?

frezbo (commented):
release note update about the spec change is v1alpha2

app/sidero-controller-manager/config/samples/metal_v1alpha2_environment.yaml
smira: yep, good point, and siderolabs, I wonder if someone is actually using those samples :shrug:

templates/cluster-template-talos-v0.13.4.yaml
smira: yep, as Sidero still supports 0.13 (we are testing with it), so we should keep the template up to date with the resources

Unix4ever (approved):

frezbo (approved):

PR #861, opened 2022-05-18
fix: append Talos default kernel args even if there is some defined

Unix4ever:
If the server or environment used by the machine had at least one kernel
param overriden, the `ipxe_server` was skipping adding all of them.

Instead of that skip only the parameters which were defined.


frezbo (approved):
good catch :+1:

smira (approved):
❤️

PR #863, opened 2022-05-18
release(v0.5.1): prepare release

Unix4ever:
This is the official v0.5.1 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,102 @@
+## [Sidero 0.5.1](https://github.com/talos-systems/sidero/releases/tag/v0.5.1) (2022-05-18)
+
+Welcome to the v0.5.1 release of Sidero!
+
+
+
+Please try out the release binaries and report any issues at
+https://github.com/talos-systems/sidero/issues.
+
+### Sidero Deployment Strategy Configuration
+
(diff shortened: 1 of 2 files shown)


hack/release.toml
```diff
@@ -6,9 +6,9 @@ github_repo = "talos-systems/sidero"
 match_deps = "^github.com/(talos-systems/[a-zA-Z0-9-]+)$"
 
 # previous release
-previous = "v0.4.0"
+previous = "v0.5.0"
 
-pre_release = true
```
smira, line 11: I think we should clean up release notes here from items which were already in 0.5.0

smira: so it lists only _new_ items

smira: so no notes, that's fine as well :)

Unix4ever: I've added some :)

CHANGELOG.md
```diff
@@ -1,281 +1,101 @@
-## [Sidero 0.5.0](https://github.com/talos-systems/sidero/releases/tag/v0.5.0) (2022-02-14)
```
smira, line 1: CHANGLOG seems to be messed up? 0.5.0 entry removed (?)

Unix4ever: oh... yeah

smira (approved):

Issue #896, opened 2022-07-07
Sidero should support hostname as Wireguard endpoint


smira:
`SIDERO_CONTROLLER_MANAGER_SIDEROLINK_ENDPOINT` should support hostnames which are resolved to IP on startup:

```
invalid Wireguard endpoint: ParseIP("sidero.dev.net"): unexpected character
```

PR #898, opened 2022-07-15
fix: resolve wireguard endpoint to IP

utkuozdemir:
If wireguard endpoint is set to be a hostname and not an IP, attempt to resolve it to an IP instead of failing immediately.

Closes siderolabs/sidero#896.

Signed-off-by: Utku Ozdemir <utku.ozdemir@siderolabs.com>


Unix4ever (approved):

smira (approved):

smira:
we should backport it and do another release!

frezbo (approved):

PR #909, opened 2022-07-21
feat: update base Talos to 1.1.1

smira:
This fixes a problem with loading config which contains new 1.1.x
fields.

Signed-off-by: Andrey Smirnov <andrey.smirnov@talos-systems.com>

PR #911, opened 2022-07-22
backports: for v0.5.3

smira:
PRs backported:

- #908
- #909 
- #910


sfyra/pkg/tests/tests.go
```diff
@@ -96,7 +96,7 @@ func Run(ctx context.Context, cluster talos.Cluster, vmSet *vm.Set, capiManager
 		},
 		{
 			"TestCompatibilityCluster",
-			TestCompatibilityCluster(ctx, metalClient, cluster, vmSet, capiManager, options.PrevTalosRelease, options.KubernetesVersion),
+			TestCompatibilityCluster(ctx, metalClient, cluster, vmSet, capiManager, options.PrevTalosRelease, "v1.22.12"),
```
frezbo, line 99: shouldn't the k8s version come from some config?

frezbo (approved):

sfyra/pkg/tests/tests.go
smira: Yes, it does, but this test specifically boots old Talos, so we need old Kubernetes

smira: so the long story was that previously we used same version for "old" Talos (0.13) and "new" Talos (0.14+). As we're right now on Talos 1.1, which is Kubernetes 1.24, it's no longer compatible with 0.13, so I have to pin old version. Hopefully we should be able to remove this test sooner or later, as it's only needed for Talos 0.13 which doesn't have SideroLink (so it's based on init nodes).

Issue #988, opened 2022-11-12
[sidero] encountered error discovering BMC IP. skipping setup: "..." rebooting in 10 seconds


vorburger:
I'm a first time user attempting to set up Sidero at home, and after much fun with PXE etc. got blocked here, it "loops" with this on the console:

    [sidero] Wipe complete
    [sidero] Attempting to automatically discover and configure BMC
    [sidero] encountered error discovering BMC IP. skipping setup: ""run /usr/sbin/ipmitool ipmitool -H -U -I open ... : Could not open device at /dev/ipmi0 or /dev/ipmi/0 or /dev/ipmidev/0: No such file or directory\n (exit status 1)" 
    [sidero] rebooting in 10 seconds

I gather that perhaps I need to somehow tell it that this (nor any of my) machine doesn't have that?

I couldn't quite figure out yet how to tell it not do that from the doc, but I'll keep looking.

In the mean time, perhaps you would like to consider handling this more gracefully, and not bail like this on machines without IPMI?

vorburger:
> I couldn't quite figure out yet how to tell it not do that from the doc, but I'll keep looking.

`export SIDERO_CONTROLLER_MANAGER_AUTO_BMC_SETUP=false` does the trick, #989 raised.

But perhaps I misunderstandood how Sidero work, as that reboot still happens, after getting rid of that error. So the reboots are not because of the "error discovering BMC IP", but because... the server isn't allocated, yet? So... onwards to https://www.sidero.dev/v0.5/getting-started/create-workload, to stop it from constantly rebooting, I guess that's the idea here; will try ASAP.

vorburger:
Let me close this issue, as I think it was a misunderstanding on my part. (#989 would still be useful to clarify that, IMHO.)

PR #989, opened 2022-11-12
docs: Detail how to disable IPMI magic (re. #988)

vorburger:
see #988


vorburger:
I've fixed some of the tpreviously failing https://github.com/siderolabs/conform checks, but for the _continuous-integration/drone/pr — Build is failing_ it looks like a maintainer needs to approve it? And don't know how to fix that _"conform/commit/gpg-identity — openpgp: signature made by unknown entity"_ ... is it documented somewhere what that actually checks how? Tx!

smira (approved):

PR #1041, opened 2023-01-31
docs: remove non-inclusive Kubernetes terms from 0.5 & 0.6 website docs

bmiddha:
Remove non-inclusive Kubernetes terms from N and N+1 website docs


smira:
Talos 0.5 and 0.6 use whatever configuration fields were back then, updating the docs won't update Talos itself.

These versions are out of support, and documentation is only kept for archive/historical reasons.

If there's anything in current (v1.3/v1.4) versions of Talos, we would accept that contribution.

PR #1144, opened 2023-05-27
Fix: code snippet lacks "sudo".

sflobbe:
When getting started, users will expect the code snippets to work as is.


smira (approved):
thank you!

PR #1163, opened 2023-06-26
docs: update taint names `master`=>`control-plane`

hxtmdev:
https://github.com/kubernetes/enhancements/blob/35befff0ad33187b2c93141d5fe1513a1b4a39a1/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint/README.md#summary

### DIFF
--- a/website/content/v0.4/Guides/bootstrapping.md
+++ b/website/content/v0.4/Guides/bootstrapping.md
@@ -138,7 +138,7 @@ You should then set your `KUBECONFIG` environment variable to the path of this f
 Because this is a single node cluster, we need to remove the "NoSchedule" taint on the node to make sure non-controlplane components can be scheduled.
 
 ```bash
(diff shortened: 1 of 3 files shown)


frezbo:
These are older version of talos shipped with K8s that did not have the name change, so not relevant.

hxtmdev:
Are you sure it should stay at the old name? I just set up a fresh cluster with sidero and had to use the new one as I was installing a current version of K8s.

Also it says `Version v0.6 of the documentation is for the Sidero Metal version being developed.`.

frezbo:
>  Also it says Version v0.6 of the documentation is for the Sidero Metal version being developed..

Sidero Uses different docs than Talos https://github.com/siderolabs/sidero/tree/main/website

hxtmdev:
I missed that the guide as a whole is targeting K8s 1.22.2, my bad. Thanks for the quick replies.

PR #1173, opened 2023-08-22
feat: update for Talos 1.5.0

smira:
Bump CAPI, controller-runtime, Talos dependencies, Kubernetes.

Fix some compatibility issues, but should be no functional changes.


Makefile
```diff
@@ -9,11 +9,11 @@ MODULE := $(shell head -1 go.mod | cut -d' ' -f2)
 
 ARTIFACTS := _out
 TEST_PKGS ?= ./...
-TALOS_RELEASE ?= v1.4.0
-DEFAULT_K8S_VERSION ?= v1.26.0
+TALOS_RELEASE ?= v1.5.0
+DEFAULT_K8S_VERSION ?= v1.27.2
```
frezbo, line 13: isn't 1.27.4 the latest?

smira: I used same as base CAPI version in this release. I think we could use almost anything, as it's just in the tests only

frezbo (approved):

PR #1178, opened 2023-09-07
docs: remove excess bash continuation backslashes

hxtmdev:
`sudo` begins a new command, so the preceding line must not end in `\`.


smira (approved):
thank you!

Issue #1212, opened 2023-10-09
Require assistance with debugging regarding PXE and the new DHCP proxy


lieberlois:
Hi everyone,

I'm trying to setup Sidero Metal for PXE Booting my SuperMicro Servers. Unfortunately, the servers dont seem to find the sidero-metal-controller during the boot process - i get a classic "pxe media test failure check cable" error. My Sidero Metal instance is running on a talos k8s cluster within the same network as the machines, i have a standard DHCP server handing out leases in my network. Both the Docker container from Talos and the Deployment from Sidero Metal are running in the host network. Using `tcpdump tcpdump -i en0 -pvn port 67 and port 68` I can also see the DHCP broadcast messages in the network on the machine, and (as a test) in a separate docker container with `--network host`. The command `curl -I http://<ip-of-talos-machine>:8081/tftp/snp.efi` also works fine from a different machine in the same network.

According to the [docs](https://www.sidero.dev/v0.6/getting-started/prereq-dhcp/#automatic-dhcp-proxy) the automatic dhcp proxy should allow me to PXE boot machines without any configuration regarding DHCP on my router  - AFAIK I don't need to configure anything (DHCP options like next-server, etc.) in my router.  Am I understanding anything wrong? The documentation doesn't really give any clues on how to debug this.

Thanks in advance guys! If we figure it out and it maybe is something missing in the docs, I am more than willing to contribute to the project :) 

Cheers,
Luis

smira:
Afaik Docker won't allow you to pass through layer 2 packets, this is a docker limitation. It can probably be worked around with some iptables tricks, but nothing I would suggest immediately.

One option is to use manual DHCP configuration for PXE boot, or run management cluster with direct access to layer 2 network.

lieberlois:
Hi @smira - thanks for the quick response. I'm trying to use Sidero Metal for my master thesis and am currently struggling with the Issue mentioned above.

1) Unfortunately I dont have control over the DHCP config in my network, which is why I wanted to try out the DHCP proxy. What would be an alternative? The docs mostly suggest creating a dockerized cluster using Talosctl - would something like k3s work?

2) Regarding Docker Limitations: I thought that using the Network "Host" basically lets the Container use the Host NIC, so layer 2 traffic should work aswell - but I might be wrong.

I would be very thankful for a small explanation to the 2 points above. If you want me to, I'll contribute to the docs so others dont fall into the same trap as I did 🤭

smira:
[...]

PR #1213, opened 2023-10-10
fix: don't send option 60 in DHCP responses

smira:
This seems to trigger the behavior to look for proxyDHCP on port 4011, see e.g.
https://www.prajwaldesai.com/pxe-e55-proxydhcp-did-not-reply-to-request-on-port-4011/.

See #1212


smira:
this is wrong, we should instead provide a way to switch off dhcp proxy

PR #1218, opened 2023-10-12
docs: Disk is double

alesti:
I think that https://www.sidero.dev/v0.6/resource-configuration/servers/#installation-disk is wrong:
```
spec:
  accepted: false
  configPatches:
    - op: replace
      path: /machine/install/disk
      value: /dev/sda
```

This leads to `machine.install.disk.disk: value`, but it needs `machine.install.disk: value`

How i found that? https://taloscommunity.slack.com/archives/CMARMBC4E/p1697122617304659

I changed it (as it is already writen in https://www.sidero.dev/v0.6/guides/patching/) and the error went away.
I found some more similar code bits and fixed them too.

### DIFF
--- a/website/content/v0.6/Guides/sidero-on-rpi4.md
+++ b/website/content/v0.6/Guides/sidero-on-rpi4.md
@@ -30,7 +30,7 @@ export SIDERO_ENDPOINT=192.168.x.x
 Generate Talos machine configuration for a single-node cluster:
 
 ```bash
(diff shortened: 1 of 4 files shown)


smira (changes requested):
the patches in question seem to be correct, and this change is not needed

```yaml
machine:
  install:
    disk: /dev/foo
```

smira:
There might be a bug somewhere else (?), but not in the updated documentation pieces

alesti:
> There might be a bug somewhere else (?), but not in the updated documentation pieces

ok, but for me it works without `/disk` in the path and i found that described in https://www.sidero.dev/v0.6/guides/patching/ -  and it does not work and shows the named unmarshalling error with `/disk`.

PR #1235, opened 2023-11-02
Handling network interface without dhcp during pxeboot

nathandotleeathpe:
High Speed Networks (e.g. HPE Slingshot)  with statically assigned IPs don't use DHCP services. This causes the ipxe `dhcp` command to fail and prevent pxe-booting from a subsequent interface. This change will allow the ipxe script to skip over these types of networks.

I've tested this change using Sidero v0.6.1. Our HSN was built using Mellonox ConnectX-5 cards on a Slingshot network.

### DIFF
--- a/app/sidero-controller-manager/internal/ipxe/ipxe_server.go
+++ b/app/sidero-controller-manager/internal/ipxe/ipxe_server.go
@@ -44,10 +44,10 @@ var ErrBootFromDisk = errors.New("boot from disk")
 var BootTemplate = template.Must(template.New("iPXE embedded").Parse(`#!ipxe
 prompt --key 0x02 --timeout 2000 Press Ctrl-B for the iPXE command line... && shell ||
 
-# print interfaces
+{{/* print interfaces */}}
 ifstat
 
-# retry 10 times overall
+{{/* retry 10 times overall */}}
 set attempts:int32 10
 set x:int32 0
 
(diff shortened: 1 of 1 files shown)


smira:
```
#25 5.904 === RUN   TestEmbeddedLength
#25 5.904     ipxe_test.go:28: 
#25 5.904         	Error Trace:	/src/app/sidero-controller-manager/internal/ipxe/ipxe_test.go:28
#25 5.904         	Error:      	"1014" is not less than "1000"
#25 5.904         	Test:       	TestEmbeddedLength
#25 5.904 --- FAIL: TestEmbeddedLength (0.00s)
#25 5.904 FAIL
```

nathandotleeathpe:
@smira

> // iPXE script should fit length of the reserved space in the iPXE binary

I'm willing to bump that test up to `1014`, but I don't have the context on how much space is left for that ipxe binary. I'm assuming there's enough extra space, since I was able to build and run the sidero deployment without issues.

smira:
> @smira
> 
> > // iPXE script should fit length of the reserved space in the iPXE binary
> 
> I'm willing to bump that test up to `1015`, but I don't have the context on how much space is left for that ipxe binary. I'm assuming there's enough extra space, since I was able to build and run the sidero deployment without issues.

We have just 1017 bytes there if I remember how things are wired correctly. (https://github.com/siderolabs/pkgs/blob/2e892fdca61391fdb060797ed372f86ca71bb5b8/ipxe/files/ipxe.script#L3-L66)

I think we should keep the size under 1000 as much as possible

app/sidero-controller-manager/internal/ipxe/ipxe_test.go
```diff
@@ -25,5 +25,5 @@ func TestEmbeddedLength(t *testing.T) {
 	}))
 
 	// iPXE script should fit length of the reserved space in the iPXE binary
-	assert.Less(t, buf.Len(), 1000)
+	assert.Less(t, buf.Len(), 1015)
```
smira, line 28: let's not bump it.. we can drop comments.. either in the code, or filter them out when generating the final script (with proper unit-tests for it)

nathandotleeathpe:
Reverted the change to the unit test. I updated the comments in the template to use text/template comments instead of ipxe comments. That will allow the template package to filter out contents in the future.

source: https://pkg.go.dev/text/template#hdr-Actions

nathandotleeathpe:
I've re-tested his change using Sidero v0.6.1.

smira:
I can't push back signed version of the commit to your branch, as it's an organization fork.

I will re-open a PR with the same commit.

PR #1240, opened 2023-11-06
fix: handle network interfaces without dhcp during pxeboot

smira:
High Speed Networks (e.g. HPE Slingshot) with statically assigned IPs don't use DHCP services. This causes the ipxe dhcp command to fail and prevent pxe-booting from a subsequent interface. This change will allow the ipxe script to skip over these types of networks.


smira:
forked from #1235

DmitriyMV (approved):

Unix4ever (approved):

frezbo (approved):

PR #1258, opened 2023-12-12
feat: update Talos to 1.6.0

smira:
Also update CAPI to 1.6.0.


app/caps-controller-manager/config/manager/manager.yaml
```diff
@@ -20,7 +20,9 @@ spec:
             - /manager
           image: controller:latest
           args:
-            - --metrics-bind-addr=127.0.0.1:8080
```
frezbo, line 23: this was dropped?

smira: yeah, CAPI moved to diagnostics-addr

frezbo (approved):

PR #1279, opened 2024-01-06
docs: Update bootstrap guide to reflect controlplane host name and stable versions

rmvangun:
I've updated the docs to reflect the use of "controlplane" vs. "master" in the bootstrapping guide. I also updated the examples to use the last stable versions of Kubernetes (1.29.0) and Talos (1.6.1).


frezbo (approved):
Thank you :heart_eyes:

PR #1288, opened 2024-01-21
fix: sidero.endpoint and sidero.mac was inverted

aarnaud:
AgentEndpointArg and AgentMACArg values was inverted


frezbo (approved):
Thank you

PR #1316, opened 2024-02-17
docs: correct "which { much => must }" typo

LukeCarrier:
Small typo I noticed when running through the Sidero install docs today.


LukeCarrier:
Thanks for the nudge to sign commits with GPG 😅

smira (approved):
thank you!

PR #1603, opened 2026-03-05
Redirect sidero.dev to siderolabs.com

tycascade:
Update `netlify.toml` to include a redirect - we're basically retiring the sidero.dev bare metal site and redirecting all its traffic to siderolabs.com

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -5,3 +5,9 @@ publish = "website/public"
 [build.environment]
 HUGO_VERSION = "0.93.2"
 NODE_VERSION = "12.21.0"
+
+[[redirects]]
+  from = "https://sidero.dev/*"
+  to = "https://siderolabs.com/:splat"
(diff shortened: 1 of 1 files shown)


tycascade:
@rothgar Justin this is the PR to redirect sidero.dev

netlify.toml
```diff
@@ -5,3 +5,9 @@ publish = "website/public"
 [build.environment]
 HUGO_VERSION = "0.93.2"
 NODE_VERSION = "12.21.0"
+
+[[redirects]]
+  from = "https://sidero.dev/*"
```
smira, line 10: I _think_ it might be easier to do this on domain level, and just drop the netlify here, @rsmitty ?

rothgar: I agree. We don't need netlify in the loop here.

PR #1605, opened 2026-03-18
fix: exclude control plane endpoint from MetalMachine addresses

scatat:
When Talos reports node addresses, it includes the shared control plane endpoint (e.g. a VIP). Sidero correctly stores these in `ServerBinding.Spec.Addresses`. But when the MetalMachine controller copies them into `MetalMachine.Status.Addresses`, the shared endpoint propagates to `Machine.Status.Addresses`, where CACPPT treats it as a per-machine Talos API endpoint. This breaks scale-down operations.

This PR filters the control plane endpoint at the CAPI boundary in `metalmachine_controller.go`, where the `Cluster` object is already in scope. `ServerBinding` retains the full address list.

Safe when `ControlPlaneEndpoint.Host` is empty — comparison against `""` matches nothing, all addresses pass through.

Ref: siderolabs/cluster-api-control-plane-provider-talos#242

🤖 Co-authored with AI

### DIFF
--- a/app/caps-controller-manager/controllers/metalmachine_controller.go
+++ b/app/caps-controller-manager/controllers/metalmachine_controller.go
@@ -211,6 +211,11 @@ func (r *MetalMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 		addresses := make([]capiv1.MachineAddress, 0, len(serverBinding.Spec.Addresses))
 		for _, addr := range serverBinding.Spec.Addresses {
+			// skip shared control plane endpoint unless it's the only address (single-node)
+			if addr == cluster.Spec.ControlPlaneEndpoint.Host && len(serverBinding.Spec.Addresses) > 1 {
+				continue
+			}
+
 			addresses = append(addresses, capiv1.MachineAddress{
 				Type:    capiv1.MachineInternalIP,
 				Address: addr,


app/caps-controller-manager/controllers/metalmachine_controller.go
```diff
@@ -211,6 +211,11 @@ func (r *MetalMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 		addresses := make([]capiv1.MachineAddress, 0, len(serverBinding.Spec.Addresses))
 		for _, addr := range serverBinding.Spec.Addresses {
+			// the control plane endpoint is shared, not per-machine
+			if addr == cluster.Spec.ControlPlaneEndpoint.Host {
```
smira: I think this has same issue as CACPPT PR in general - this might break valid setups with single node controlplane.

StephenTan-TW: Good catch — fixed in f3d428b.

Added `&& len(serverBinding.Spec.Addresses) > 1` so the filter only fires when the machine has other addresses. Single-node CP where the endpoint is the only address keeps it.

StephenTan-TW:
Squashed to a single commit with DCO sign-off. The GPG identity check will fail since I'm not in the siderolabs org — that one needs a maintainer override.

app/caps-controller-manager/controllers/metalmachine_controller.go
smira: I still don't think it's correct, it's trying to guess something while the context is missing here.

Either it should be fixed on Talos side to stop reporting VIPs, or Sidero Metal should pull addresses itself.

